PR #632, opened 2023-03-12
feat(vue): add component toast

TylerAPfledderer:
Adds the `Toast` component for Vue

- [ ] fix issue where the template ref for `ToastGroup` is lost after created toasts are removed.


codebender828 (approved):
Thanks @TylerAPfledderer for teh initial work here

PR #637, opened 2023-03-14
feat(vue): add component environment

TylerAPfledderer:
Add the `Environment` component for Vue. (Does not include Docs)


codebender828 (approved):
Thanks @TylerAPfledderer for the prior work here

PR #677, opened 2023-03-26
fix(vue/checkbox): update state handling

TylerAPfledderer:
PR to improve passing of state from Zag machine.

- remove slot props for input and label and pass props through the parent checkbox slot
- Add test for indeterminate
- Use `checked` as a named v-model


TylerAPfledderer:
@codebender828 noting here that the issue I commented above might resolve itself here once I resolve current conflicts and get the branch up-to-date

codebender828 (approved):
@TylerAPfledderer Thanks for the prior work done on this bad boi

Issue #727, opened 2023-04-05
Nuxt 3 / SSR support


lewebsimple:
Hello, great project !!

I'm trying this out in a Nuxt 3 project and it seems that some components only work on the client-side (Dialog for example). The error is `ReferenceError: Element is not defined`

I'm not sure if this is preventable or not.

Shyrro:
Hey @lewebsimple , thanks for reporting this issue 🙏 

It should now be fixed in the latest rc : `yarn add @ark-ui/vue@rc`. 

Let us know if you notice another issue, thanks !

lewebsimple:
Stupendous, it works like a charm !

Any plans for including a nuxt 3 module ?

PR #733, opened 2023-04-07
fix(ssr): replace Element prop type with Object

Shyrro:
The type "Element" was breaking SSR for the dialog.

The context of the dialog was not reactive so i wrapped it inside a reactive to be able to spread it without losing reactivity.

Fixes issue : (#727)

### DIFF
--- a/packages/vue/src/dialog/use-dialog.ts
+++ b/packages/vue/src/dialog/use-dialog.ts
@@ -1,19 +1,19 @@
 import { connect, machine, type Context } from '@zag-js/dialog'
 import { normalizeProps, useMachine } from '@zag-js/vue'
-import { computed } from 'vue'
-import { transformComposableProps, useId } from '../utils'
+import { computed, reactive } from 'vue'
+import { useId } from '../utils'
 
 export type UseDialogProps = {
   context: Omit<Context, 'id'>
   emit: CallableFunction
 }
 
-export const useDialog = (props: UseDialogProps) => {
-  const { context, emit } = transformComposableProps(props)
+export const useDialog = (emit: CallableFunction, context: Omit<Context, 'id'>) => {
+  const reactiveContext = reactive(context)
 
   const [state, send] = useMachine(
     machine({
-      ...context,
+      ...reactiveContext,
       id: useId().value,
       onClose() {
         emit('close')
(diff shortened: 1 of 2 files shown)


packages/vue/src/dialog/use-dialog.ts
```diff
@@ -1,19 +1,19 @@
...
+import { computed, reactive } from 'vue'
+import { useId } from '../utils'
 
 export type UseDialogProps = {
   context: Omit<Context, 'id'>
   emit: CallableFunction
 }
 
-export const useDialog = (props: UseDialogProps) => {
-  const { context, emit } = transformComposableProps(props)
```
TylerAPfledderer, line 12: @Shyrro `transformComposableProps` was supposed to return a reactive context and still be able to destructure.

TylerAPfledderer (commented):
Hey @Shyrro! Quick thought here. 😄

packages/vue/src/dialog/use-dialog.ts
Shyrro: Hey @TylerAPfledderer !
The transform composable does this : 

```
export function transformComposableProps<T extends { context: object }>(props: T): T {
  return {
    ...props,
    context: reactive(props.context),
  }
}
```

The object returned is not `reactive`. The context is. This means you can destructure context, not the resulting object. And destructuring the result will result in the context losing its reactivity.

For this to work, all the resulting object should be wrapped in the `reactive` function. Also, inside that function, `props` shouldn't be destructured to spread them. if you want to spread them, you should do `...reactive(props)`.

Hope that answers your question !

TylerAPfledderer: @Shyrro understood!

With the refactor you did for the dialog composable, I'm finding this function unnecessary. The point of it was to simplify a pattern that was occuring.

I think that it should be removed for all composables.

And overall I would say that `defaultValue` would not be a prop to apply for some of the composables if `modelValue` is the preferred choice. (Since the ref applied to `modelValue` can take an initial or default value)

So each composable would have at least `emit` and `context` props and applied in the way you have done in this PR.

What do you think?

Shyrro: @TylerAPfledderer , since we talked about this in Slack i'm merging this ;) 
We'll do the refactor later

PR #834, opened 2023-04-28
feat(vue): add color-picker component

TylerAPfledderer:
Adds the ColorPicker component for vue

### DIFF
--- a/.changeset/gold-trains-rest.md
+++ b/.changeset/gold-trains-rest.md
@@ -1,0 +1,5 @@
+---
+'@ark-ui/vue': minor
+---
+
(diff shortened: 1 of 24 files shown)


TylerAPfledderer:
Hey @cschroeter, is it OK to go ahead and merge this despite the export check failing? I've added an export type `ColorPickerContext` to accommodate for the lack of inferred typing when exposing the api. 

Everything else should be there. (Minus prop type for the trigger because it is not needed in Vue)

cschroeter:
I fixed the `exportCheck` script. you should be good to go :)

PR #866, opened 2023-05-03
fix(vue): make `useId` isomorphic

### LINKED ISSUE #825: [Vue]: Unique ID doesn't work in SSR
Hey again,

Stumbled on an issue while using various components. The unique id created with the `useId()` utility within the component and passed to the machine isn't unique when using SSR. In addition, any `id` prop passed to the component is ignored. Doing a brief scan, it looks like some components are also still missing the `id` prop.

As of now, it seems the only solutions are either:

1. Wrap every component in `<ClientOnly>`.

2. Develop an SPA.

These reproductions show some issues stemming from this. In the `Select` repro, the second menu position is not where it should be. In the `Editable` repro, the active styling remains on the second input when focusing between the two inputs. In either one, wrapping the components in `ClientOnly` or adding `ssr: false` to the `nuxt.config.ts` should fix the issues and create a different id as expected (which you can see in the DOM).

https://stackblitz.com/edit/nuxt-starter-xzqjnc?file=app.vue

https://stackblitz.com/edit/nuxt-starter-g6mate?file=nuxt.config.ts

Hope this helps. Perhaps there's another solution I'm unaware of as well.

codebender828:
Makes `useId` hook for Vue isomorphic, such that it renders a unique id on server as well as client.

Further changes need to be made to allow the components to accept an id prop, which the `useId(id: string)` can accept.

Closes #825


Shyrro (approved):
Good with me, i think you can merge this.

Small note for later: I'm wondering if we can create a singleton array, that would contain created ids, so that we could use it to generate a "unique" id in SSR?

PR #904, opened 2023-05-07
feat(docs): website main navigation hover event

sumi-0011:
## as is
수정 전전

## to be
수정 후 메인

--- 

When I hovered main navigation in existing docs, I saw that there was no action other than the cursor changing, so I thought it would be bad for UX. 

So I changed the color of the link by adding css, which changes the color of the link when I hovered the navigation link, and improved the UX.


cschroeter:
@sumi-0011 Thanks you for your contribution

PR #906, opened 2023-05-07
fix(docs): TableOfContent scrollspy doesn't work

sjoleee:
There is a problem that when a page is scrolling, the appearance of the `Link` component inside the `TableOfContent` component does not change automatically.

This is because the `useScrollSpy` hook is not working properly. Instead of getting `"#" + item.slug` injected, we got `item.slug` injected and added `"#"` inside the hook.

Also, I assigned `item.slug` as the ID of Link, so that the `querySelector` works properly.

### asis
ark_asis

### tobe
ark_tobe


cschroeter:
Thanks for your contribution <3

Issue #905, opened 2023-05-07
[Menu] [Vue]: slot props


jiblett1000:
### Description

Just like was done for the Select component https://github.com/chakra-ui/ark/pull/831

Please expose slot props for `isOpen` and the `highlightedItem` (or whatever it should be named) state.

### Proposed Solution or API

N/A

TylerAPfledderer:
Hey @jiblett1000 !

Thank you for pointing this out and the indirect reminder about exposing the api! I'll make this my next assignment to look over all the components to ensure the related api's are exposed to the `v-slot` where needed.

We might still be considering the addition of using the `expose` method for added flexibility. :)

jiblett1000:
@TylerAPfledderer All sounds great! definitely adding/using expose couldn't hurt :smile:

jiblett1000:
@TylerAPfledderer BTW, I guess I'll just wait for https://github.com/chakra-ui/ark/issues/913 to resolve before posting more slot prop related issues. Seems like it might be kind of redundant at this point. :thinking:

ploca14:
I think the MenuItem should expose isHighlighted through slot props.

PR #918, opened 2023-05-08
fix(vue): add id props and improve state exposition

### LINKED ISSUE #825: [Vue]: Unique ID doesn't work in SSR

### LINKED ISSUE #913: [Vue]: Allow for exposure of machine API to `v-slot` for needed components
### Description

Part of an overall cleanup with the Vue package is being able to expose the individual component's machine API through the `v-slot` so that state such as boolean checks or selected values can be accessed by the user to then apply to custom structures within each component's wrapper.

If not immediately, considering to also apply the `expose` method with the APIs for added flexibility.

### Additional Information

Related to #905

Shyrro:
BREAKING CHANGE: Update exposed context types and remove unnecessary expose calls

This PR contains :

- Addition of id prop to all components ( Closes #825 )
- Update to types to be as close as possible to the machine
- Addition fo slot props for exposure (update slot props when needed and expose when needed) ( Closes #913 , #905 )

### DIFF
--- a/packages/vue/src/accordion/accordion.tsx
+++ b/packages/vue/src/accordion/accordion.tsx
@@ -1,45 +1,46 @@
 import { type Context } from '@zag-js/accordion'
 import { defineComponent, type PropType } from 'vue'
 import { ark, type HTMLArkProps } from '../factory'
-import { type Assign } from '../types'
+import { type Assign, type Optional } from '../types'
 import { createVueProps, type ComponentWithProps } from '../utils'
 import { AccordionProvider } from './accordion-context'
 import { useAccordion } from './use-accordion'
 
 export type AccordionContext = Context & { modelValue?: AccordionContext['value'] }
-export type AccordionProps = Assign<HTMLArkProps<'div'>, AccordionContext>
 
-const VueAccordionProps = createVueProps<AccordionProps>({
+export type UseAccordionProps = Assign<HTMLArkProps<'div'>, AccordionContext>
+
+const VueAccordionProps = createVueProps<UseAccordionProps>({
   id: {
-    type: String as PropType<AccordionProps['id']>,
+    type: String as PropType<UseAccordionProps['id']>,
   },
   modelValue: {
-    type: [String, Object] as PropType<AccordionProps['modelValue']>,
+    type: [String, Object] as PropType<UseAccordionProps['modelValue']>,
   },
   collapsible: {
-    type: Boolean as PropType<AccordionProps['collapsible']>,
+    type: Boolean as PropType<UseAccordionProps['collapsible']>,
     default: false,
   },
   multiple: {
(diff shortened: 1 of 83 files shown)


cschroeter:
```json
"id": {
    "type": "string",
    "isRequired": true,
    "description": "The unique identifier of the machine."
  },
```

The generated types, in this case the accordion are not 100% correct. The `id` should be optional. We achieve this in React / Solid doing sth like this:

```tsx
export type UseAccordionProps = Optional<accordion.Context, 'id'>
```

jiblett1000:
Hey ya'll. Just wondering if this PR is taking into account the recent release of Vue 3.3 and specifically being able to use imported types in SFC? I just upgraded and the exported types from Ark don't seem to be compatible for some reason. 

Just curious as this could be a huge time saver as I wouldn't need to re-write all of the props in my custom components, but just import an extend the ones from Ark instead.

Shyrro:
Hey @jiblett1000 

This PR is not related, as the SFC types are only available under SFC's. We use TSX to build our Vue components so we don't have the same issues as we can already use Typescript how we want. 

The exported types are pure typescript so you can reuse them in your own `defineProps` if needed. 

Could you elaborate a little bit more on what you mean by "exportes types don't seem compatible for some reason", what errors do you have etc?

Thanks

jiblett1000:
Hey @Shyrro I'm happy to create an issue for this if need be. This is where I'm importing / using the type:

```ts
import type { NumberInputProps } from "@ark-ui/vue";

interface Props {
  label: string;
  id?: string;
  paddingRight?: number;
  paddingLeft?: number;
}

const props = defineProps<NumberInputProps & Props>();
```

I modified an existing stackblitz to show the error.... although, I'm getting a different error there than I am locally. 

https://stackblitz.com/edit/nuxt-starter-3jzcrb?file=components%2FChildDialog.vue

PR #947, opened 2023-05-12
feat/vue as child

codebender828:
Implements `asChild` prop for Vue. Replaces the polymorphic factory with native proxy.

### In Vue

```vue
<dialog-trigger as-child>
  <span>Dialog Trigger (asChild &rarr; Vue SFC)</span>
</dialog-trigger>

<!-- renders <span /> element -->
```

### In TSX

```tsx
<DialogTrigger id={id} asChild class="space-inline-end-4">
  <span>Dialog Trigger (asChild &rarr; Vue JSX)</span>
</DialogTrigger>

//  renders <span /> element
```

**Changes:**
- feat: create as-child for vue
- docs(vue): add documentation for as child


packages/vue/src/utils.ts
```diff
@@ -89,3 +90,109 @@ export function transformComposableProps<T extends { context: object }>(props: T
...
+  // Do not bind any event handlers when there is a `disabled` or `aria-disabled` prop set.
+  if (target.disabled || target['aria-disabled']) {
+    return Object.assign(
+      target,
+      // Set all event listeners that we collected to `undefined` so that `cloneVNode` from Vue doesn't drop them. `cloneVNode` only merges the
+      // existing and new props, they don't just override therefore we have to
+      // explicitly declare and nullify them.
+      Object.fromEntries(Object.keys(eventHandlers).map((eventName) => [eventName, undefined])),
+    )
+  }
```
segunadebayo, line 178: This is not an assuming we make in React or Vue. I'd rather not have us to this because it's too implicit.

packages/vue/src/utils.ts
```diff
@@ -89,3 +90,109 @@ export function transformComposableProps<T extends { context: object }>(props: T
...
+    return [child]
+  })
+}
+
+/**
+ * We use our own `mergeProps` instead of Vue's mergeProps to correctly handle event listeners.
+ *
+ * Credit: headless-ui/vue: https://github.com/tailwindlabs/headlessui/blob/main/packages/@headlessui-vue/src/utils/render.ts#L196
+ */
+export function mergeProps(...propsArgs: Record<any, any>[]) {
```
segunadebayo, line 145: We also want to merge styles and classnames applied to the elements. Please see the merge-props we use in React: https://github.com/chakra-ui/zag/blob/main/packages/core/src/merge-props.ts

packages/vue/src/factory.tsx
```diff
@@ -2,18 +2,124 @@
...
+            }
+          }
+          return cloned
+        }
+
+        if (Array.isArray(children) && children.length === 1) {
+          return children[0]
+        }
+
+        return children
```
Shyrro, line 90: So if i understand correctly, if there are no attributes on the element calling `asChild` we allow more than one element?

packages/vue/src/utils.ts
```diff
@@ -89,3 +90,109 @@ export function transformComposableProps<T extends { context: object }>(props: T
...
+  // No children
+  if (input == null) return false
+  // Raw HTML elements like 'div', 'span', e.t.c.
+  if (typeof input.type === 'string') return true
+  // Other Vue component objects
+  if (typeof input.type === 'object') return true
+  // Functional components
+  if (typeof input.type === 'function') return true
+  // HTML Comment nodes or Text nodes
+  return false
```
Shyrro, line 107: Maybe ?

```suggestion
  return input && (typeof input.type === 'string' || typeof input.type === 'object' || typeof input.type === 'function')
```

Just a suggestion, i actually like the comments on each condition

packages/vue/src/utils.ts
```diff
@@ -89,3 +90,109 @@ export function transformComposableProps<T extends { context: object }>(props: T
...
+  }
+
+  // Merge event handlers
+  for (const eventName in eventHandlers) {
+    Object.assign(target, {
+      [eventName](event: { defaultPrevented: boolean }, ...args: any[]) {
+        const handlers = eventHandlers[eventName]
+
+        for (const handler of handlers) {
+          if (event instanceof Event && event.defaultPrevented) {
```
Shyrro, line 187: Didn't know we could use `instanceof` in a js condition !

Shyrro (approved):
Huge work here @codebender828 , props to you !

packages/vue/src/utils.ts
segunadebayo: We probably want to change that check to use a more generic expression like “preventDefault in event”.

This instanceof check won’t work within custom window environments.

packages/vue/src/factory.tsx
```diff
@@ -2,18 +2,138 @@
...
+        let children = slots.default?.()
+        children = renderSlotFragments(children || [])
+
+        if (Object.keys(attrs).length > 0) {
+          const [firstChild, ...otherChildren] = children
+          if (!isValidVNodeElement(firstChild) || otherChildren.length > 0) {
+            const componentName = instance?.parent?.type.name
+              ? `<${instance.parent.type.name} />`
+              : 'component'
+            throw new Error(
```
codebender828, line 74: Improved error message.

PR #984, opened 2023-05-21
refactor(vue): update machine context handling

TylerAPfledderer:
Update context handling for the Vue component machines to match the other framework packages, to allow the machine to handle updates when the props change.

Also cleans up the docs, types, and includes the `defaultValue` props where missing.

### Fixing context passed to the machine
Per usage for React and Solid `useMachine` has two arguments. The first argument is the context object holding all the initial values to set up the machine. The second argument takes an object of options, which includes a `context` object of the props with any that have updated their value in order for the machine to watch and update internally when necessary.


segunadebayo:
Chris and I worked to improve the context handling based on the design of the machine. Thanks for taking a stab at this.

Issue #1142, opened 2023-07-05
Custom interactions with Toast via descriptions


martinpengellyphillips:
### Description

Enable richer interaction with toasts by supporting JSX for description or supporting custom render (as in Zag).
For example, a toast with a link in the description is useful to allow users to click on the toast to take them to the relevant item in the app.

### Proposed Solution or API

Support https://zagjs.com/components/solid/toast#custom-render-functions in Ark.

Or

Support JSX elements for `description` rather than only strings.

cschroeter:
Thanks for taking your time to report this issue. A fix is on the way

PR #1150, opened 2023-07-06
fix: resolve an issue with custom rendered toasts

cschroeter:
closes Custom interactions with Toast via descriptions #1142

### DIFF
--- a/packages/react/src/toast/toast.stories.tsx
+++ b/packages/react/src/toast/toast.stories.tsx
@@ -83,9 +83,10 @@ const ExampleComponent = () => {
       <button
         onClick={() => {
           toast.create({
-            render: () => (
+            title: 'Data submitted!',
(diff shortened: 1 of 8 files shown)


packages/react/src/toast/toast.stories.tsx
```diff
@@ -72,13 +72,27 @@ const ExampleComponent = () => {
...
             placement: 'bottom-start',
           })
         }}
       >
         Add bottom-start toast
       </button>
+      <button
+        onClick={() => {
+          toast.create({
+            render: () => (
```
martinpengellyphillips, line 86: Does the custom `render` receive `props` as in Zag?

cschroeter: Yes it does

cschroeter:
Closes #1150

PR #1166, opened 2023-07-10
feat: add CarouselIndicatorGroup and CarouselIndicator parts

renatodeleao:
Hello again 👋 

Was building a custom carousel on top of akr’s one and noticed that the machine implementation [from `@zagjs`](https://github.com/chakra-ui/zag/blob/a4f6369976d426c537625aa17457dbe7c6fe54cb/packages/machines/carousel/src/carousel.connect.ts#L169-L183) has support for two extra parts: `indicator` and `indicatorGroup`, which I believe that in the carousel context are typically that bullet navigation overlay with the slide counts. I decided to bring it them into `ark` as it is a fairly common _piece of the puzzle_ and a requirement for my implementation. 

### Note for reviewers
1. Solid version is untested, don't have enough experience with the framework and with my current work, I don't have more time to dedicate.
1. By default, the indicator doesn't have the `onClick` event handler in their props, so I've manually added it using `scrollTo` exposed method to make the bullets clickable. It seemed a sensible default but consumers might want to disable this behaviour via prop so I'm not 100% sure.
1. I'm also not 100% that website docs are websites are correct for `react` because locally running the site seemed not to pick up the updates I've made to the story but hopefully will be working in production, please double-check/edit 😅  

### Final notes

I'm somewhat of an experienced developer and It's not super easy to contribute to this package.
- I came to add a feature on `vue`, but end up having to port it to `react`, because if I want to document it I have to build the react version as the website demos are all react under the hood.
   - There are no contributing guidelines saying what I need to update to make it happen — again, easy to figure out for experienced devs, hard for beginners.
   - The styling of parts of the demo has a bit of boilerplate associated, hard for people not acquainted with `panda` or `css-in-js` in general as it's one more thing to learn to contribute. 

Please don't take this the wrong way, I'm so glad Ark UI exists and what it represents for the Vue ecosystem!  I understand that this package is relatively new to have all these things figured out, I'm just saying this because I really want `ark` to become the most popular headless UI library and for that, it needs to grab the max amount of contributions possible and that's also done by eliminating these barriers.

Thank you so much for all the work on Ark, keep shipping :shipit:


cschroeter:
First of all, thank you for your thoughtful contribution and the in-depth description of your changes. Your commitment to improving Ark UI is highly appreciated.

I want to reassure you that I'll be reviewing your PR shortly. I understand your concern about some aspects of the contribution process. Your feedback is very valuable to us as we continuously strive to make the process as clear and accessible as possible.

> To update the stories you need to run `pnpm document` in the root folder (I just did that for you).

cschroeter:
Thanks for your contribution <3

PR #1176, opened 2023-07-13
fix: script prepare-release typo

iamdin:
fix script prepare-release typo, change `pakcageName` to `packageName`


cschroeter:
Haha good catch ;)

PR #1253, opened 2023-08-07
fix(vue): [Carousel] CarouselControl wasn't returning a component and…

Baben01:
Hello, 

While using the `Carousel` component in Vue, I came across a minor issue in the documentation and found a bug within the `CarouselControl` component. The bug can be observed in the following code snippet:

```tsx
return () => {
  return () => [Component]
}
```

Additionally, I noticed a couple of minor things in the documentation for the Carousel component:
 - The MDX syntax used for `Vue` code examples was set as `html` or `tsx`, which caused some issues with syntax highlighting.
 - The documentation did not provide sufficient information about using the `CarouselControl` component in `Vue` compared to the details available for `React` documentation.
 
 Let me know if you have question or comments :D 
 
Thanks a lot for this library.

### DIFF
--- a/packages/vue/CHANGELOG.md
+++ b/packages/vue/CHANGELOG.md
@@ -6,6 +6,10 @@ description: All notable changes to this project will be documented in this file
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed a bug in the `CarouselControl` component where it was returning a function instead of the expected component. Use the `CarouselControl` component in the `Vue` documentation just like in `React`.
+
 ## [0.6.0] - 2023-08-06
 
 ### Added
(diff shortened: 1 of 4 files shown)


cschroeter:
Thanks for creating this PR. I just found out  there was another issue with this carousel:

```tsx
import { mergeProps } from '@zag-js/vue'
import { computed, defineComponent } from 'vue'
import { ark, type HTMLArkProps } from '../factory'
import { useCarouselContext } from './carousel-context'
import { carouselAnatomy } from './carousel.anatomy'

export type CarouselControlProps = HTMLArkProps<'div'>

export const CarouselControl = defineComponent({
  name: 'CarouselControl',
  setup(_, { slots, attrs }) {
    const api = useCarouselContext()

    const mergedProps = computed(() =>
      mergeProps(api.value.nextTriggerProps, attrs, carouselAnatomy.build().control.attrs),
    )

    return () => {
      return () => <ark.div {...mergedProps.value}>{slots.default?.()}</ark.div>
    }
  },
})
```

```api.value.nextTriggerProps``` ???
I've updated this on the zag-js 0.14.0 branch. thanks for taking care of this <3

PR #1369, opened 2023-09-04
[vue][Slider] fix: hiddenInputProps not being spread correctly rendering hidden input to the DOM

renatodeleao:
## TL;DR
- the computed property `hiddenProps` needs to be spread by `.value`
- the computed property was actually redundant because the `modelValue` is not bindable to a `input` element, so I've removed it and used `api.value.hiddenInputProps` directly in template.
- added basic spec (because docs demos are written in react, this bug was visually caught)

## Details

- I only caught this because I have a "smoke-snapshot" test on my Ark-ui component wrappers, and my CI broke when upgrading to `ark/vue@0.7.0`
- [Reproduction example](https://stackblitz.com/edit/vitejs-vite-r3j91c?file=src%2FApp.vue&terminal=dev)

Screenshot 2023-09-04 at 12 40 29

Screenshot 2023-09-04 at 12 45 20

### DIFF
--- a/packages/vue/src/slider/slider-thumb.tsx
+++ b/packages/vue/src/slider/slider-thumb.tsx
@@ -10,17 +10,12 @@ export const SliderThumb: ComponentWithProps<SliderThumbProps> = defineComponent
   setup(_, { slots, attrs }) {
     const api = useSliderContext()
 
-    const hiddenInputProps = computed(() => ({
-      ...api.value.hiddenInputProps,
-      modelValue: api.value.value,
-    }))
-
     return () => (
       <>
         <ark.div {...api.value.thumbProps} {...attrs}>
           {slots.default?.()}
         </ark.div>
-        <input {...hiddenInputProps} />
+        <input {...api.value.hiddenInputProps} />
       </>
     )
   },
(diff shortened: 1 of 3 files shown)


packages/vue/src/slider/slider-thumb.tsx
```diff
@@ -10,17 +10,12 @@ export const SliderThumb: ComponentWithProps<SliderThumbProps> = defineComponent
   setup(_, { slots, attrs }) {
     const api = useSliderContext()
 
-    const hiddenInputProps = computed(() => ({
-      ...api.value.hiddenInputProps,
-      modelValue: api.value.value,
-    }))
```
renatodeleao, line 16: **question:** hiddenInputProps doesn't seem to include `api.value.value` so we might need to use a computed property after all if we want the input to have the current modelValue binding.


```diff
const hiddenInputProps = computed(() => ({
    ...api.value.hiddenInputProps,
-    modelValue: api.value.value,
+    value: api.value.value,
  }))
```

renatodeleao:
@cschroeter decide to do a fix for this one directly — as it's pretty straightforward — without creating an issue first but let me know, when you have the time, if the contribution process should have been different.

Cheers.

cschroeter:
@renatodeleao thanks for your contribution. it is highly appreciated. i'm currently in the pase of revisiting all vue components to bring them on the same level as their solid and react counterparts.

I will make sure that your fix will be part of this <3

Issue #1405, opened 2023-09-13
[Solid] The asChild attribuite doesn't work as expected 


navaru:
### Description

The `asChild` prop, set on the parent, overwrites the child's `class` and `style` attributes.

### Link to Reproduction

https://stackblitz.com/edit/vitejs-vite-z8khkc?file=src%2FApp.tsx

### Steps to reproduce

Check the `<CheckboxLabel>` element, add/remove the `asChild` attribute. The `span` class is replaced with the class of the parent.

### Ark UI Version

0.11.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

PR #1406, opened 2023-09-13
fix(solid): asChild spread issue

navaru:
Fixes parent and child attributes merging when using `asChild` prop.

Resolves: https://github.com/chakra-ui/ark/issues/1405


cschroeter:
You are fk rockstar. Report an issue, explain the issue,  file a PR with(!) some tests. I have no words but thank you.

Issue #1455, opened 2023-09-22
[solid][datepicker] lag when selecting date in range mode


ryoid:
### Description

When using `DatePicker` in "range" mode, after selecting the first date and hovering other dates there is major lag.

Inspection of dev tools show that many elements are updating - I guess these unnecessary updates are causing the lag.

### Link to Reproduction

https://stackblitz.com/edit/gapu-solid-query-avzbtm?file=src%2Fcomponents%2FDateRangePicker.tsx

### Steps to reproduce

1. Open range DatePicker
2. Open dev tools and expand html element tree for datepicker cells
3. Click once (select first date of range)
4. Hover around other dates, should be able to feel the lag and see the updates in dev tools

### Ark UI Version

0.11.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

### Browser

Microsoft Edge Version 117.0.2045.31

### Additional Information

https://github.com/chakra-ui/ark/assets/18044529/8237d3af-b0cd-44a0-9698-7cd635b21b80

ryoid:
> Took a quick look. It seems Solid.js is re-creating all the DOM nodes when the state changes, which is undesirable.
> 
> A new unique id is created each time you interact with the date picker.

Adding on incase it helps, this also happens in the [Ark storybook](https://github.com/chakra-ui/ark/blob/main/packages/frameworks/solid/src/date-picker/date-picker.stories.tsx#L77) as well. 

You can observe the updates when you move the mouse around after clicking a cell.

```tsx
<DatePicker selectionMode="range" numOfMonths={2}>
  {(api) => {
    createEffect(() => {
      console.log(api().weeks)
    })
    return <>...</>
  }}
</DatePicker>
```

ryoid:
Found a solution, using `<Index>` instead of `<For>` prevents the updates to cause DOM updates.

Opened a PR to update the storybook example https://github.com/chakra-ui/ark/pull/1456

EfstathiadisD:
[](url)This issue is present in React as well. Also there is another issue here. When we move in either Solid or React the cursor from one month to the next the Months start moving. I have attached a Video. I think we need some example of how to construct a Dual Calendar. (i.e: The Docs state that we should put numOfMonths={2} which isn't correct as then the offset treats both months as a set.)

https://github.com/chakra-ui/ark/assets/48406873/71dca961-bbf8-43b0-9454-09246553b927

[...]

PR #1456, opened 2023-09-22
fix: solid datepicker

ryoid:
Current Solid DatePicker storybook has performance issues https://github.com/chakra-ui/ark/issues/1455

Using `<Index>` instead of `<For>` for the cells prevents the updates to cause DOM updates, leading to performance issues.

---

I also noticed that the `DatePickerGrid` id calls `createUniqueId()` that updates in the same way, not sure if it is intended behavior. I believe this does not have to keep updating?

https://github.com/chakra-ui/ark/blob/main/packages/frameworks/solid/src/date-picker/date-picker-grid.tsx#L17


segunadebayo:
Correct. We should probably move the createUniqueId call for the grid outside the memo to avoid recalling it.

PR #1564, opened 2023-10-20
feat(react): file upload

Omikorin:
### Notes
`FileUpload.ItemPreview` returns `null` if Zag has thrown an error because we cannot render a preview of a file type other than `image/`


Omikorin:
done

PR #1576, opened 2023-10-24
feat: improve a11y for disclosure components

cschroeter:


packages/website/src/content/types/react/combobox.types.json
```diff
@@ -211,6 +211,10 @@
       "isRequired": false,
       "description": "Render as a different element type."
     },
+    "fallback": {
```
segunadebayo, line 214: We probably want to omit `fallback` from the typedocs

PR #1604, opened 2023-10-30
refactor: use for instead of deprecated htmlFor

Omikorin:
This needs work in Zag types to get compatibility.


cschroeter:
`for` is a reserved keyword so `htmlFor` isn't that bad of a choice. however when we can make this happen for solid so lets go

PR #1661, opened 2023-11-09
fix(menu): add presence api to positioner component

### LINKED ISSUE #1660: [React][Menu] Presence is not working on the positioner component
### Description

Hi!

The `Menu.Positioner` component is always mounted even if we set the `lazyMount` property to `true` on the `Menu.Root` component.

If we look at the [source code](https://github.com/chakra-ui/ark/blob/main/packages/frameworks/react/src/menu/menu-positioner.tsx), we can see that the presence logic is missing:

```ts
const presenceApi = usePresenceContext()

if (presenceApi.isUnmounted) {
  return null
}
```

### Link to Reproduction

https://codesandbox.io/s/purple-feather-gwcqsw?file=/src/Menu.tsx

### Steps to reproduce

1. Look at the devtools
2. Check the positioner element in the DOM: it's mounted

### Ark UI Version

1.0.0

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

### Browser

Google Chrome 119

### Additional Information

_No response_

fseglard:
Closes #1660

### DIFF
--- a/packages/frameworks/react/src/menu/menu-positioner.tsx
+++ b/packages/frameworks/react/src/menu/menu-positioner.tsx
@@ -1,6 +1,7 @@
 import { mergeProps } from '@zag-js/react'
 import { forwardRef } from 'react'
 import { ark, type HTMLArkProps } from '../factory'
+import { usePresenceContext } from '../presence'
 import { useMenuContext } from './menu-context'
 import { type UseMenuReturn } from './use-menu'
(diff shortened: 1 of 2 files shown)


cschroeter:
@fseglard 

Thanks for your contribution. I took your examples but also had to updated solid and a few more things.
But I set you as co-author <3

PR #1665, opened 2023-11-09
fix(dialog): resolve an issue where dialog would not animate on exit

cschroeter:
For some reason the dialog content ends with `hidden: false` 

```js
Object { ref: composeRefs(node)
, "data-state": "closed", hidden: true }
[popover-content.tsx:18:2](http://localhost:6006/src/popover/popover-content.tsx)
```

```js
Object { ref: composeRefs(node)
, "data-state": "closed", hidden: false }
[dialog-content.tsx:18:2](http://localhost:6006/src/dialog/dialog-content.tsx)
```


segunadebayo:
I just fixed this from Zag.js [259f358](https://github.com/chakra-ui/zag/commit/259f3584d257de7c811dbb4b031fb563f2a7976d). The positioner had the hidden attribute applied to it

PR #1754, opened 2023-11-21
docs: add overflowX=auto to remove overflow on api reference on mobile

juliomuhlbauer:
Screenshot_20231121_111220_Chrome


juliomuhlbauer:
Also, path.join doesn't work as expected on Windows. Because the path becomes docs\\overview instead of docs/overview.

Omikorin:
Please look at this [https://nodejs.org/api/path.html#pathposix]
This should be a standardized solution for your problem.

Omikorin:
I removed node-specific imports because of issues related to [Vite](https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility).

Great work @juliomuhlbauer, thank you!

PR #1772, opened 2023-11-26
docs: add missing props in slider examples

koba04:
The `Slider.Thumb` component, which is used as an example in the [Getting Started page](https://ark-ui.com/docs/overview/getting-started), requires the `index` prop, so the example gets a type error. This PR adds the `index` prop to the example.

https://ark-ui.com/docs/components/slider

image


Omikorin:
Thank you

koba04:
Thank you!

PR #1795, opened 2023-11-29
docs: fix keyboard navigation in Menu component

alajusticia:
I was playing with the example for the Menu component in the docs, and I realized that some items were unreachable when navigating with the keyboard and pressing their first letter.

It was caused by extra spaces in the item content. So this little PR fixes that by removing the extra spaces.


Omikorin:
Thank you!

PR #1898, opened 2023-12-15
fix(vue): tooltip controlled, scoped slots & stories

iamdin:
- fixed open controll state
- fixed scoped slots context
- pref stories cases


Omikorin:
Thank you @iamdin! :)

PR #1911, opened 2023-12-19
fix(vue/accordion-item): set default for `disabled` prop to `undefined`

TylerAPfledderer:
This PR addresses the issue where setting the `disabled` prop for `Accordion.Root` to true was not propagating to the items.

In `Accordion.Item`, the `disabled` prop was declared with a default value of `false`, perpetually taking precedent when checking item state. This is because `getItemState` in the [Zag Accordion Connect](https://github.com/chakra-ui/zag/blob/305ac564f33c05a209d403dc2008984cc7520e17/packages/machines/accordion/src/accordion.connect.ts#L25) is running a nullish coalescing for `isDisabled` to obtain its value .

The default value still needs to be declared, but it should be `undefined` so the nullish coalescing returns false in the check and obtains that context's disabled value instead.

Also, remove the skip for the related test.

### DIFF
--- a/packages/frameworks/vue/src/accordion/accordion-item.tsx
+++ b/packages/frameworks/vue/src/accordion/accordion-item.tsx
@@ -15,7 +15,7 @@ export const AccordionItem = defineComponent({
     },
     disabled: {
       type: Boolean,
-      default: false,
+      default: undefined,
     },
   },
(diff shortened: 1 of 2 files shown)


Omikorin:
Thanks mate :)

PR #1927, opened 2023-12-23
refactor(vue): revise select, stories & test

iamdin:


Omikorin:
Let's use that new API for defineComponent. You were right and we can achieve two things at once

iamdin:
@Omikorin That's great！I will refactor in the next commit. By the way，does that mean we will refactoring all components with the new way？

Omikorin:
Yeah, we'll do it for all the components. With [Vue 2 EOL](https://blog.vuejs.org/posts/vue-2-eol), I think there will be more users using min. v3.3 that we need anyway.

I'll commit here one change to Presence to show you something.

Omikorin:
Okay, please look at #1785. Files: `accordion.tsx`, `accordion-item.tsx`, `accordion-item-content.tsx`.

We changed a method for using presence from `<Presence>` to `usePresence` / `usePresenceProps` and API directly in the component

iamdin:
Thank you so much for the showcase！

Omikorin:
I hope Function Signature won't break anything in the meantime hah

iamdin:
@Omikorin  Thank you! I've been busy with work recently, will be back later

PR #1939, opened 2023-12-30
fix(vue): presence onExitComplete & types

iamdin:
- fix `onExitComplete` cannot call correctly
- simplify the define of props
- fix type use `toValue`

### DIFF
--- a/packages/frameworks/vue/src/presence/presence.props.ts
+++ b/packages/frameworks/vue/src/presence/presence.props.ts
@@ -1,3 +1,5 @@
+import { declareEmits } from '../utils'
+
 export const props = {
   present: {
     type: Boolean,
@@ -6,3 +8,5 @@ export const props = {
   lazyMount: Boolean,
   unmountOnExit: Boolean,
 }
+
+export const emits = declareEmits(['exit-complete'])
--- a/packages/frameworks/vue/src/presence/presence.props.ts
+++ b/packages/frameworks/vue/src/presence/presence.props.ts
@@ -1,21 +1,8 @@
-import type { Context } from '@zag-js/presence'
-import type { PropType } from 'vue'
-import { declareEmits } from '../utils'
-import type { UsePresenceProps } from './use-presence'
-
 export const props = {
   present: {
-    type: Boolean as PropType<Context['present']>,
+    type: Boolean,
     default: undefined,
   },
-  lazyMount: {
-    type: Boolean as PropType<UsePresenceProps['lazyMount']>,
-    default: false,
-  },
-  unmountOnExit: {
-    type: Boolean as PropType<UsePresenceProps['unmountOnExit']>,
-    default: false,
-  },
+  lazyMount: Boolean,
+  unmountOnExit: Boolean,
 }
-
-export const emits = declareEmits(['exit-complete'])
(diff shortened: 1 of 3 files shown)


packages/frameworks/vue/src/presence/presence.props.ts
```diff
@@ -1,21 +1,8 @@
...
-  },
-  unmountOnExit: {
-    type: Boolean as PropType<UsePresenceProps['unmountOnExit']>,
-    default: false,
-  },
+  lazyMount: Boolean,
+  unmountOnExit: Boolean,
 }
-
-export const emits = declareEmits(['exit-complete'])
```
iamdin, line 21: As we use Function Signature to define component, I don't think we need to declare emit to get the type hint of event, because they already exist in the definition of zag

packages/frameworks/vue/src/presence/use-presence.ts
```diff
@@ -18,20 +18,19 @@ export interface UsePresenceProps extends Optional<presence.Context, 'present'>
...
+      emit('exit-complete')
+    },
+  }))
   const wasEverPresent = ref(false)
   const nodeRef = ref<VNodeRef | null>(null)
 
-  const [state, send] = useMachine(
-    presence.machine({
-      ...context.value,
-      onExitComplete: () => {
```
iamdin, line 29: If zag/presence set initial context during the initialization phase, we can back this.

packages/frameworks/vue/src/presence/presence.tsx
```diff
@@ -19,20 +19,6 @@ export const Presence = defineComponent<PresenceProps>(
...
-      lazyMount: {
-        type: Boolean as PropType<UsePresenceProps['lazyMount']>,
-        default: undefined,
-      },
-      unmountOnExit: {
-        type: Boolean as PropType<UsePresenceProps['unmountOnExit']>,
-        default: undefined,
-      },
-    },
-    emits,
```
iamdin, line 36: As we use Function Signature to define component, I don't think we need to declare emit to get the type hint of event, because they already exist in the definition of zag

Omikorin: Good work with cleaning `props` in the component.

I'm not sure if `emits` are needed in components

packages/frameworks/vue/src/presence/presence.props.ts
```diff
@@ -1,21 +1,12 @@
...
-  lazyMount: {
-    type: Boolean as PropType<UsePresenceProps['lazyMount']>,
-    default: false,
-  },
-  unmountOnExit: {
-    type: Boolean as PropType<UsePresenceProps['unmountOnExit']>,
-    default: false,
-  },
+  lazyMount: Boolean,
+  unmountOnExit: Boolean,
```
Omikorin, line 9: `as PropType...` should be here, simply because `x.props.ts` is a generated (`pnpm scripts vue:props`) file with good typing. It tells you what is bad in a more clean way.
I tried 
image
but this is not clean as the old method.
image

With the old method:
image

Omikorin:
@iamdin I've reverted what should be reverted.

Thanks!

iamdin:
> @iamdin I've reverted what should be reverted.
> 
> 
> 
> Thanks!

Thanks！

PR #1943, opened 2023-12-30
fix(vue): components presence

Omikorin:
If checked then done/fixed etc

Not working lazy mounting (from tests):

- [x] Combobox
- [x] Hover Card
- [x] Popover
- [x] Tabs

Aria errors (tests):

- [x] Dialog.Trigger aria-controls
- [x] Menu.Trigger aria-controls
- [x] Popover.Trigger aria-controls

Presence existing in component (rewritten):

- [x] Accordion
- [x] Color Picker
- [x] Combobox
- [x] Date Picker
- [x] Dialog (flaky tests, presence didn't work from the start)
- [x] Hover Card
- [x] Menu
- [x] Popover
- [x] Select
- [x] Tabs
- [x] Tooltip

Note: Working Presence is understood as `lazyMount` passed to a component's `Root` part (like in Chakra UI `isLazy`). There is no reason to prefer props in items (I know there is, although it's a simpler and more intuitive approach).


Omikorin:
omg it's working

Omikorin:
@TylerAPfledderer @iamdin I'd be happy for your review. I hope I didn't miss anything important.

Maybe it's that last PR before Ark UI 1.0 for Vue ^^

TylerAPfledderer:
@Omikorin any reason why react and solid tests are showing up here?

Omikorin:
> @Omikorin any reason why changes in react and solid tests are showing up here?

Yes, during work I found flaky or missing tests between those, so I synced them by the rule "do it now while you remember".

packages/frameworks/vue/src/dialog/dialog-trigger.tsx
```diff
@@ -1,15 +1,24 @@
...
 export interface DialogTriggerProps extends HTMLArkProps<'button'> {}
 
 export const DialogTrigger = defineComponent<DialogTriggerProps>(
   (_, { slots, attrs }) => {
     const api = useDialogContext()
+    const presenceApi = usePresenceContext()
+
+    const triggerProps = computed(() => ({
+      ...api.value.triggerProps,
+      'aria-controls': presenceApi.value.isUnmounted
```
TylerAPfledderer, line 15: I don't think this attribute should disappear after the initial rendering of the dialog content with `lazyMount`. Once the dialog content is rendered to the DOM, this attribute should stay attached to the trigger, else the purpose of it is lost.

In other words, the attribute only matters when a user navigates to the trigger, and it should be present if the associated content is visible in the DOM.

I also acknowledge that this attribute [does not have much support with screen readers at the moment](https://a11ysupport.io/tech/aria/aria-controls_attribute); it will not be acknowledged by any of the major SRs, with only JAWS ensuring the user's ability to jump to the controlled element.

My above comments stand in regards to future-proof.

packages/frameworks/vue/src/menu/menu-trigger.tsx
```diff
@@ -1,15 +1,24 @@
...
 export interface MenuTriggerProps extends HTMLArkProps<'button'> {}
 
 export const MenuTrigger = defineComponent<MenuTriggerProps>(
   (_, { slots, attrs }) => {
     const api = useMenuContext()
+    const presenceApi = usePresenceContext()
+
+    const triggerProps = computed(() => ({
+      ...api.value.triggerProps,
+      'aria-controls': presenceApi.value.isUnmounted
```
TylerAPfledderer, line 15: Same comment as with `DialogTrigger`

packages/frameworks/vue/src/popover/popover-trigger.tsx
```diff
@@ -1,15 +1,24 @@
...
 export interface PopoverTriggerProps extends HTMLArkProps<'button'> {}
 
 export const PopoverTrigger = defineComponent<PopoverTriggerProps>(
   (_, { slots, attrs }) => {
     const api = usePopoverContext()
+    const presenceApi = usePresenceContext()
+
+    const triggerProps = computed(() => ({
+      ...api.value.triggerProps,
+      'aria-controls': presenceApi.value.isUnmounted
```
TylerAPfledderer, line 15: Ditto comment with `DialogTrigger`

TylerAPfledderer (dismissed):
@Omikorin on initial pass, this is great and I do not see anything missing related to presence.

I only have one tiny nitpick though 😄

packages/frameworks/vue/src/dialog/dialog-trigger.tsx
Omikorin: Thanks, I've checked the behavior. This attribute disappears only when using `lazyMount` with `unmountOnExit` (like in `Dialog` `LazyMount` story). Otherwise, it stays in place along with lazy-mounted content.

It's the same situation in React, so let's ask Chris.

@cschroeter, what do you think about a11y case that Tyler brought?

cschroeter: @TylerAPfledderer @Omikorin 

I just glimpsed over the conversation. `lazyMount` or `unmountOnExit` has often a negative impact on a11y. It is a tradeoff for performance. My recommendation is to avoid using `lazyMount` or `unmountOnExit` if possible.

I've read a bit about this topic here: https://github.com/w3c/aria/issues/995 and as far as I can tell there is no clear winner if `aria-controls` should be removed, empty or just point to the hidden element. 

Under this assumption please keep the current solution as it passes the Storybook a11y Addon and this axe addon for jest some contributor brought up. 

Cheeers

Omikorin: Thanks mate. I agree with the current solution.

@TylerAPfledderer I'll dismiss the request then and merge PR. I appreciate yourreview! :)

L-orz:
The `Combobox` close animation is not displaying correctly because `presenceProps` is not being passed to `Combobox.Content`.
https://github.com/chakra-ui/ark/blob/b5e025723d71364a75d873a4af710c587d949348/packages/frameworks/vue/src/combobox/combobox-content.tsx#L17

PR #1971, opened 2024-01-02
fix(vue/factory): expose intrinsic attr props from `HTMLPolymorphicProps`

### LINKED ISSUE #1884: `HTMLPolymorphicProps` type not exposing intrinsic element attributes
### Description

The type signature of [HTMLPolymorphicProps](https://github.com/chakra-ui/ark/blob/4f8d697dc07f1a760d069781cdd8b3ffe3f79850/packages/frameworks/vue/src/factory.tsx#L51-L53) for the Vue package seems to be incomplete, as it is not properly extracting props from `IntrinsicElementAttributes` of a given HTML tag.

This effects prop typing of components that extend `HTMLArkProps`

N/A

### Steps to Reproduce

You can view in any component story in package the lack of auto completion of props that come from `InstrinsicElementAttribute` of a given HTML tag. For example, if `HTMLArkProps<"img">` you will not get the auto-completion of props such as `src`, `alt`, or any of the `aria-*`

### Ark UI Version

0.9.0

### Framework

- [ ] React
- [ ] Solid
- [X] Vue

### Browser

N/A

### Additional Information

There is already a possible solution that is inline with this type's usage in the other frameworks (the other frameworks do not actually have the `HTMLPolymorphicProps` type and just `HTMLArkProps`, but the general approach is the same to work in Vue).

For the generic type supplied to `HTMLPolymorphicProps`, it is only extending `keyof IntrinsicElementAttributes`, which means `ExtractPropTypes` is receiving a string type and not an object type, thereby extracting nothing.

The fix:
```ts
export type HTMLPolymorphicProps<T extends ElementType> = Omit<
  // If T is the keyof a DOM element (i.e. `img`) then need to obtain that element's attribute object to extract props.
  // Otherwise, return T which could be a component (`DefineComponent` definition)
  ExtractPropTypes<T extends DOMElements ? IntrinsicElementAttributes[T] : T>,
  'ref'
> & {
  asChild?: boolean
}
```
## Warning 🚨
This issue is only created for tracking and should be considered backlogged at this time. Currently, the caveat to the above fix creates the type error `Type instantiation is excessively deep and possibly infinite.` related to [ComponentWithProps](https://github.com/chakra-ui/ark/blob/4f8d697dc07f1a760d069781cdd8b3ffe3f79850/packages/frameworks/vue/src/utils.ts#L33-L41).

There is a consideration in doing away with `ComponentWithProps` in favor of receiving the necessary typing from the [defineComponent function signature approach](https://github.com/vuejs/core/blob/a6503e3e77b04a1ab8c07078e07091b06d51131d/packages/runtime-core/src/apiDefineComponent.ts#L97-L130) introduced to aid in user-defined generic typing of components.

Unless there is a solution for the instantiation error that I could not recognize, need to wait until determining if the change to `defineComponent` is viable and be completed first.

TylerAPfledderer:
Allow for intrinsic attributes to be exposed as props through `HTMLPolymorphicProps`. Previously they were not being extracted properly.

With the update in `defineComponent` to be used in the function signature approach, this change does not result in the error `Type instantiation is excessively deep and possibly infinite` as it did with the previous approach.

Also, this PR cleans up some typing errors associated with props exposed.

Closes #1884


Omikorin:
Nice. Do we know anything about the function signature approach in generic components? ( in our case `Combobox.Root` and `Select.Root`)

TylerAPfledderer:
> Nice. Do we know anything about the function signature approach in generic components? (in our case `Combobox.Root` and `Select.Root`)

I'm still stumped on this one. I would think creating a wrapper to house the type of generic we are using, but then the typing gets lost when trying to render the component and use its props along with slots and emits.

That article I showed in the related issue was basically that with a backwards approach to what we are trying to achieve (because it was a solution they offered prior to the function signature coming online). So if anything I would try and see if that works here so we can expose the dynamic prop types correctly.

Omikorin:
I think if we merge #2011 then we could merge this without fixing generics for now

PR #1972, opened 2024-01-02
fix(vue): usePresence types

iamdin:
I think this might be a way to solve the usePresence type


packages/frameworks/vue/src/presence/use-presence.ts
```diff
@@ -18,7 +18,7 @@ export interface UsePresenceProps extends Optional<presence.Context, 'present'>
 
 export type UsePresenceReturn = ReturnType<typeof usePresence>
 
-export const usePresence = (props: UsePresenceProps, emit: CallableFunction) => {
+export const usePresence = (props: MaybeRef<UsePresenceProps>, emit: CallableFunction) => {
```
Omikorin, line 21: I guess it works

Omikorin (approved):

Omikorin:
Thank you!

PR #1977, opened 2024-01-03
fix(vue/collection): reactivity with collection

iamdin:
In component Combobox, reactivity can be lost by directly deconstruct props. So I try to fix this by create `collection` in `context`, also Select.

### DIFF
--- a/packages/frameworks/vue/src/combobox/use-combobox.ts
+++ b/packages/frameworks/vue/src/combobox/use-combobox.ts
@@ -21,23 +21,20 @@ export const useCombobox = <T extends CollectionItem>(
 ): UseComboboxReturn<T> => {
   const getRootNode = useEnvironmentContext()
 
-  const { items, itemToString, itemToValue, isItemDisabled, ...comboboxProps } = props
   const context = computed(() => {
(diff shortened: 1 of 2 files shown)


Omikorin:
It's an interesting question why in the select we have a watch and in the combobox we don't.

iamdin:
Using `watch` to subscribe `props.items` was the way I originally thought of solving the reactivity issues, and it works.
But when i in-depth understanding `useMachine` in`zagjs`, and I found this discussion https://github.com/chakra-ui/zag/discussions/994#discussioncomment-7510199. I realized that using `context` is a better way to sync/update the machine's context at runtime. 

Secondly, we have subscribed to `props` in the `computed` function at the beginning (for `context`). If we still use watch to subscribe `props.items`, there will be twice reactivity callbacks when `props.items` changed.

Hope you can understand what i mean. @Omikorin 

And thanks for the merged.

Omikorin:
Okay, I understand! I am not a Vue person, so it was like magic for me why we were doing this when machines should do the work.
Thanks for the explanation @iamdin :)

PR #1980, opened 2024-01-04
refactor(vue/checkbox): use v-model:checked for two-way binding & reactivity context

iamdin:
- refactor `v-model` to `v-model:checked` for two-way binding
- use `useMachine` context to solve reactivity issues


packages/frameworks/vue/src/checkbox/checkbox.props.ts
```diff
@@ -5,6 +5,7 @@ import { declareEmits } from '../utils'
 export const props = {
   checked: {
     type: String as PropType<Context['checked']>,
```
Omikorin, line 7: The `checked` type can be the same as `defaultChecked` according to the same context. Are there any typing errors there?

packages/frameworks/vue/src/checkbox/stories/checkbox.story.vue
```diff
@@ -1,64 +1,60 @@
...
       </Checkbox.Root>
     </Variant>
 
     <Variant title="Indeterminate">
       <div style="padding-left: 20px; display: flex; flex-direction: column; gap: 4px">
-        <Checkbox.Root v-model="parentChecked">
-          <Checkbox.Control data-testid="parent-control">
-            <span v-if="parentChecked === true">✓</span>
-            <span v-else-if="parentChecked === 'indeterminate'">-</span>
+        <Checkbox.Root v-model:checked="parentChecked" v-slot="{ isChecked, isIndeterminate }">
```
Omikorin, line 41: nice

packages/frameworks/vue/src/checkbox/checkbox.props.ts
iamdin: Oh yes, I missed it. Because we used type assertion `PropType<Context['checked']>`, so there is no typing error.

Omikorin:
Thanks

PR #1982, opened 2024-01-05
refactor(vue/dialog): use v-model:open for two-way binding

iamdin:


packages/frameworks/vue/src/dialog/use-dialog.ts
```diff
@@ -1,28 +1,33 @@
...
   const getRootNode = useEnvironmentContext()
-  const reactiveContext = reactive(props)
 
   const [state, send] = useMachine(
     dialog.machine({
-      ...reactiveContext,
-      id: reactiveContext.id || useId().value,
+      ...context.value,
+      id: context.value.id || useId().value,
+      open: props.open ?? props.defaultOpen,
```
Omikorin, line 26: I found this in the Checkbox, should we change it to `checked: props.checked ?? props.defaultChecked` too?

image

packages/frameworks/vue/src/dialog/stories/dialog.story.vue
```diff
@@ -9,7 +9,7 @@ const open = ref(false)
 <template>
   <Story title="Dialog">
     <Variant title="Basic">
-      <Dialog.Root>
+      <Dialog.Root default-open>
```
Omikorin, line 12:
```suggestion
      <Dialog.Root>
```
To sync with other stories

Omikorin (approved):

Omikorin:
Thank you very much!

packages/frameworks/vue/src/dialog/use-dialog.ts
iamdin: yep, i missed it

Omikorin: I'll fix that

PR #2004, opened 2024-01-10
fix(tags-input): fix issue with tags-input

cschroeter:


cschroeter:
Fixed in the latest release.

Issue #2007, opened 2024-01-12
[solid] problem with `asChild` prop and svgs


arekmaz:
### Description

When I wrap an `svg` element with e.g. `<ark.div class="text-sm" asChild>` I expect the class prop to be passed down and composed with existing svg class, like it works for other elements.

Instead the class is not passed down, I think the issue is in this line:
https://github.com/chakra-ui/ark/blob/main/packages/frameworks/solid/src/clone-element.ts#L11

`SVGElement` is not an instance of `HTMLElement`, I think the check should be extended also to `SVGElement`, or even better to `Element` which should contain every possible type of html tag

### Link to Reproduction (or Detailed Explanation)

just create the elements, the provided screenshots should be enough to show the issue

### Steps to Reproduce

![image](https://github.com/chakra-ui/ark/assets/24407822/652dcfbc-22c8-4d2d-97d5-025445629e59)
![image](https://github.com/chakra-ui/ark/assets/24407822/b74ab2a9-d3ad-474f-a89d-2ea9e635f0ba)
![image](https://github.com/chakra-ui/ark/assets/24407822/beb674ee-f214-4741-95a8-9ef6f2040ada)
![image](https://github.com/chakra-ui/ark/assets/24407822/d926e01c-cb05-46b7-bc93-db26eefc71b9)

### Ark UI Version

1.2.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

### Browser

ARC (Chromium)

cschroeter:
@arekmaz 

Thanks for taking your time to create this issue. I can confirm the issue:

<img width="161" alt="grafik" src="https://github.com/chakra-ui/ark/assets/1846056/c519113f-3110-406a-9dff-043e643d113e">

cschroeter:
I managed to re-create this [issue](https://github.com/chakra-ui/ark/blob/41e6cd44dcdaf214b559852c355126c47d7e812f/packages/frameworks/solid/src/factory.test.tsx#L98)

arekmaz:
I have a fix for this svg thing, but it works only for the initial render, after updates it falls apart

I think #2008 has to be solved first to meaningfully resolve this issue

arekmaz:
https://github.com/chakra-ui/ark/pull/2009

PR #2009, opened 2024-01-12
fix(solid): make `asChild` prop work with svg children

arekmaz:
this mr allows svgs to be passed as children to components with `asChild`, before the prop spread was disabled as svgs are not html elements

references #2007


cschroeter:
@arekmaz

Thanks for taking your time to create this PR. I've slightly modified your PR and also udpated the changelog.

Issue #2008, opened 2024-01-12
[solid] `asChild` prop issues after component updates


arekmaz:
### Description

in my fork of the repo I was trying to provide a fix for #2007, I wrote the following test: (in the solid `spread.test.ts` file)
![image](https://github.com/chakra-ui/ark/assets/24407822/a4abad1e-4683-4eaa-b9c6-93818aa83d89)
![image](https://github.com/chakra-ui/ark/assets/24407822/895532a8-0c3f-4b6f-9e65-a540e3f2216a)

during the initial render the parent class is correctly merged into the child, however after updates (clicks), it looks like the parent class is no longer present

![image](https://github.com/chakra-ui/ark/assets/24407822/ae4bddd4-7c51-49fe-9678-213984220346)

I tried to investigate how is the `asChild` implemented, I know that the `asChild` topic for solid is a bit more tricky than e.g. in react, but I'm too new to solid

is this behavior expected but perhaps uncocumented, or is there a special way of using the `asChild` prop with solid, or is this just a bug and otherwise the `asChild` api is figured out?

### Link to Reproduction (or Detailed Explanation)

https://codesandbox.io/p/devbox/blissful-butterfly-mkysr3?file=%2Fsrc%2FApp.tsx

### Steps to Reproduce

1. go to repro link
2. click on the "bounce" button

### Ark UI Version

1.2.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

### Browser

ARC (Chromium)

cschroeter:
@arekmaz 

I started to look into this issue for ohter reasons (hydration missmatch).
This `spread.test.ts` is a bit strange. I started to mimic the React `factory.test.tsx`.

Here is what I have so far:

```tsx
import { render, screen } from '@solidjs/testing-library'
import user from '@testing-library/user-event'
import { vi } from 'vitest'
import { ark } from './factory'

const ComponentUnderTest = () => (
  <ark.div
    id="parent"
    data-part="parent"
    data-testid="parent"
    class="parent"
    style={{ background: 'red' }}
    asChild
  >
    <ark.span
      id="child"
      data-part="child"
      data-testid="child"
      class="child"
      style={{ color: 'blue' }}
    >
      Ark UI
    </ark.span>
  </ark.div>
)

describe('Ark Factory', () => {
  it('should render only the child', () => {
    render(() => <ComponentUnderTest />)

    expect(() => screen.getByTestId('parent')).toThrow()
[...]

PR #2010, opened 2024-01-14
[solid] introduce Kobalte inspired `As` component to properly merge props with `asChild` prop

arekmaz:
# this would be a breaking change if accepted

this is meant as a prototype/proposal for changes, some tests are missing, some forked parts could be duplicated in ark, e.g. `styleObjectFromString`, but I want at least to see if the idea is valid

## Description:

When I issued this https://github.com/chakra-ui/ark/issues/2008, I noticed that the `asChild` mechanism in `ark-ui/solid` is not working correctly, basically it was working only for the first render

it merged the parent props with the current dom attributes of the child, due to `props.children` being a plain DOM node (no `props.children.props` like in react), the problem is, that the child attributes after update are already a result of merging

I did some digging and the two solutions I found:
 - render props:
   ```tsx
   <P asChild>{props => <C {...mergeProps(props, {...})} />}</P>
   ```
  - the `As` component inspired by https://kobalte.dev/:
    ```tsx
    <P {...parentProps} asChild><As component={C} {...childProps} /></P>
    ```
    which exposes child props to the parent (much like in react) and allows for proper merge
    
this PR is a proposal to adopt the `As` component as well as working implementation,
    
much of the stuff is forked from kobalte and solid primitives,
    
I added a test which checks if the merging works correctly after the updates (a similar test was failing before this changes), and it passes


cschroeter:
@arekmaz 

Thanks for taking your time to create this proposal / PR. I will have a chat with @segunadebayo  later this week.

cschroeter:
Closing this in favour of https://github.com/chakra-ui/ark/pull/2050

PR #2011, opened 2024-01-15
fix(vue): select and combobox root

Omikorin:
I managed to get something like:

```tsx
export const Select = defineComponent(
  (props: SelectProps<CollectionItem>, { slots, attrs, emit }) => {
```

but it's worse and infers emits to the return type.

Trying to get generics done isn't working because of `any` errors about `slots`, `attrs` and `emit`.

My workaround returns what we want, but with one bug - now it is possible to use `string` along with `object` (I think users won't make this mistake, the console will show errors). Yeah, so it's working better than earlier


Omikorin:
@TylerAPfledderer I've made a workaround to generics for now. It's simple and works better without serious issues.

Tell me if it makes sense to merge

TylerAPfledderer (approved):
@Omikorin this is better.

The main issue to try and overcome from here is with the runtime props.

Expectation is that if the `items` prop receives `{label: string, value: string}[]`, then the param in `itemToString` say should infer `(item: { label: string, value: string }) => something`. This currently is not happening.

However, I think we should go ahead and push this PR to maintain consistency with the other components using the function signature. And if we revert anything or when we modify further, we can better address the changes to all components.

Leave no component behind, no? 😅

Omikorin:
Yeah, that's my reasoning. In addition, if we move to SFC then we'll rewrite these components too.

PR #2032, opened 2024-01-15
feat(solid):  replace `asChild` with `as` prop

cschroeter:
This is a breaking change for the next major release.
Here is what the proposal looks like:

_Before:_

```tsx
export const Basic = () => (
  <Popover.Root>
    <Popover.Trigger asChild>
       <Button>Open me</Button>
    </Popover.Trigger>
  </Popover.Root>
)
```

After:

```tsx
export const Basic = () => (
  <Popover.Root>
    <Popover.Trigger as={Button}>
       Open me
    </Popover.Trigger>
  </Popover.Root>
)
```


cschroeter:
Closing in favour of https://github.com/chakra-ui/ark/pull/2050

PR #2030, opened 2024-01-15
fix(solid): added reactivity to collection and context of use-select (#2029)

crasite:
This PR fixed #2029 
- The `use-select` hook currently doesn't update the state machine of `zag-js` when `collectionOptions`'s props change.
- This PR make that hook register change in `collectionOptions`'s props including `items` props.
- Which in turn solve #2029

Issue #2029, opened 2024-01-15
Select's Value Text can't display dynamically update items


crasite:
### Description

Newly added item to `items` via signal doesn't get displayed properly

### Link to Reproduction (or Detailed Explanation)

https://stackblitz.com/edit/solidjs-templates-xozmyg?file=src%2FApp.tsx

### Steps to Reproduce

1. create a signal for items
2. use said signal as `items` props on `Select.Root`
3. add new item to that signal
4. select that new item

_**Default Placeholder is shown instead of the new item**_

### Ark UI Version

1.2.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

### Browser

*

Reverier-Xu:
Hi, this issue still happened on the latest rc `3.0.0-2`, see https://stackblitz.com/edit/solidjs-templates-numplr?file=package.json,src%2FApp.tsx

PR #2039, opened 2024-01-16
fix(solid): add reactivity to collection and context of use-combobox

crasite:
This PR apply to same fix from #2030 to the `Combobox` component.

The same bug (#2029) also appears in this component ([StackBlitz](https://stackblitz.com/edit/solidjs-templates-teegxc?file=src%2FApp.tsx))


Omikorin:
Thank you :)

PR #2058, opened 2024-01-21
test(react): improve performance

Omikorin:
## Reason

Tests `should render part %s` are meaningful and provide more info than a single `should render` test. The tradeoff is performance because Jest automatically takes care of the cleanup. In the worst-case scenario, we end up with 22 renders of Date Picker. CI would run faster with this change.

Approx. 1.5 s for each render
image

## Proposed solution

- Disable [global cleanup](https://testing-library.com/docs/react-testing-library/setup/#skipping-auto-cleanup) (for a single test - by importing from `@testing-library/react/pure`)
- Describe the `parts & exports` set of tests with one cleanup after **all** tests
- Enable cleanup after **each** test in the second set of "normal" tests

## Other solutions
- Remove parts test and check only one render (not meaningful)

## Effect
After the fix, Date Picker tests take only 30% of time.

image


cschroeter:
@Omikorin 

LGTM. Any downsides to this appraoch?

Omikorin:
> @Omikorin
> 
> LGTM. Any downsides to this appraoch?

I didn't find any

PR #2078, opened 2024-01-23
feat(factory): add `as` prop

Zyr0nX:
Add the ability to use `as` prop with ark factory, this will allow creating reusable components like `<Heading />` when you're not sure which tag to use


segunadebayo:
Hi @Zyr0nX,

We don't intend to support this `as` prop in VDOM-related frameworks. The `asChild` prop is sufficient for most use cases.

```jsx
<Typography asChild>
   <h4>Some Heading</h4>
</Typography>
```

Alternatively, you can build a simple component like this:

```jsx
const Heading = ({ as: Component = "h3", ...rest }) => <Component {...rest} />
```

PR #2107, opened 2024-01-29
feat: documenting data-attributes poc

FilipePfluck:
Ark-ui comes with several data-attributes which are quite handy. However, they are not well-documented, so it is a bit hard for the user to figure it out. I think we should include those on the documentation. 

This pull request is an attemp to do that. To achieve this, I declared new types for the Accordion component data-attributes. If those types are already declared somewhere, we could use them instead. 

Keep in mind this is just a POC and not meant to be merged, so I commented some things to get it working. Also, most likely the data-attributes declarations should be made in zag and not in ark, but I'm still creating this PR as a poc. I also didn't botter in organizing the different data-attributes in separate files.

One last consideration regarding UX: the way my POC is right now, the data attributes look like they are a Component Part of their own. I think we could display the data-attributes a bit differently so it is clear for the user that they are the data-attributes of each Component Part.

### DIFF
--- a/packages/frameworks/vue/CHANGELOG.md
+++ b/packages/frameworks/vue/CHANGELOG.md
@@ -44,6 +44,7 @@ description: All notable changes to this project will be documented in this file
 
 ### Fixed
 
+- Fixed an issue where emitted event caused "not declared" warning
 - Fixed an issue on touch devices where selecting an item within `Combobox`, `Menu`, or `Select` triggered a click event on the element behind the portalled content.
 - Fixed an issue in `PinInput` where pasting a value filled all inputs instead of populating them one per input.
 - Fix issue where `Select` component submits its first option when used in a form, even if there is no value selected.


FilipePfluck:
@Omikorin you might be interested in checking this out

Omikorin:
I took a glimpse at it. I guess exposed `Context` is a bug caused by testing this poc. The API reference on the website is good, but your idea to separate them is the best.

Our target is to have data attributes from one place for all frameworks i.e. from an anatomy or Zag.
Let's expose these `data attrs` types from their roots - Zag. By this, we can improve Zag's docs and instantly update attributes whenever needed.

What do you think @segunadebayo?

FilipePfluck:
> I took a glimpse at it. I guess exposed `Context` is a bug caused by testing this poc. The API reference on the website is good, but your idea to separate them is the best.
> 
> Our target is to not expose them unnecessarily from components and have data attributes from one place for all frameworks i.e. from an anatomy or Zag. Let's expose these `data attrs` types from their roots - Zag. By this, we can improve Zag's docs and instantly update attributes whenever needed.
> 
> What do you think @segunadebayo?

The exposed context is because I commented a line where we were filtering to show only things ending with `Props`. Some proper validation would do the trick. 

And I agree documenting those in zag is the rigth way to go. The rest of the types definitions already comes from there

Omikorin:
@FilipePfluck I am back with the decision.

Our team decided this is too complex and a simple table of data attributes or some sort of styling guide that shows the attributes in use would help.

Let's start simple first

PR #2116, opened 2024-02-02
feat: add svelte setup

Omikorin:
As always, new problems. Storybook works when downgrading to Svelte 4 but we need v5 for Zag integration.

## Blockers

- [ ] storybookjs/addon-svelte-csf#156


JReinhold:
👋 Storybook maintainer here, love your work.

If you're okay with writing stories in regular CSF instead of using the Svelte CSF addon, you should be able to use an upcoming beta version of Storybook 8. I have a branch of Storybook 8 working with Svelte 5, should be releasable soon.

I can't promise anything in regards to the addon as it's community maintained.

Omikorin:
@JReinhold thank you! I'll be waiting for the update :)

It'll be great to be able to use Storybook because setting a build for a library properly is a tricky part here 😅 

I believe in the Svelte CSF addon. It's a great possibility to write stories closer the Svelte-y approach. It'll help generating code snippets for docs too

JReinhold:
Storybook version `8.0.0-rc.2` should add support for Svelte version `5.0.0-next.65` - most likely also later versions than that but no guarantees given it's still prerelease.

https://github.com/storybookjs/storybook/releases/tag/v8.0.0-rc.2

`npx storybook@next upgrade`

🙂

algoflows:
Anyone know how we can use slots without using svelte CSF addon? We're stuck in limbo as can't build many of our components as most rely on slots to work and it seems CSF isn't yet supported....

JReinhold:
> Anyone know how we can use slots **without** using svelte CSF addon? We're stuck in limbo as can't build many of our components as most rely on slots to work and it seems CSF isn't yet supported....

You can use a middle-layer component (still needs a better name) to translate args to slots. Depending on your docs needs, it could look something like this: https://stackblitz.com/edit/github-mbaaau?file=src%2FButton.stories.ts

Instead of using the real component in your stories (in this case `Button`) you create another component (`ButtonStoriesHelper`) that takes in props and adds them as slots to the real component. If you need stories to show off more than just string content in slots, you have to define those components separately.

I know this is not very ergonomic unfortunately, but it's the only way to do this since Svelte doesn't have an API for passing in slots in pure JS, the only way to pass slots to a component is from within a `.svelte` file.

PR #2121, opened 2024-02-02
docs: documenting data-attributes

FilipePfluck:
This pull request adds documentation for the data-attributes of the Accordion component. Please tell me if this format is ok and if I can do the same thing for the rest of the components. Also let me know if I missed any data-attribute


FilipePfluck:
I'll be doing this bit by bit whenever I get some time. In the meanwhile I think the components I already documented can be reviewed, I may have missed some data-attributes. If I got something wrong go ahead and correct me.

segunadebayo:
Thanks for working on this.

After reviewing this closely, I've decided to automate it with a script in Zag.js. See the code here: 9b379db27efa01207772551d95995a28d9ad206f

This will make sure it's always in sync with the code.

PR #2149, opened 2024-02-06
chore: move to bun

cschroeter:


Omikorin:
I need the SvelteKit template

Omikorin:
> @Omikorin
> 
> The installation failed because of some eslint plugin for svelte. Why do we have an example project if we don't even have Svelte? I am confused :D

I've explained it on Slack.

Storybook doesn't support Svelte 5 yet, so a few options remain. #2116

cschroeter:
To much effort

Issue #1787, opened 2023-11-28
Using Portal produces a TypeScript error


aureldussauge:
### Description

I'm using `Portal` inside a `Popover` but it produces an error. I have to add `// @ts-ignore` to make it work

<img width="1023" alt="image" src="https://github.com/chakra-ui/ark/assets/12541829/8fb7b3dd-c77b-4dd9-9548-aef7cb8cabc8">

### Link to Reproduction (or Detailed Explanation)

https://codesandbox.io/p/devbox/hardcore-frog-kfkk9c?file=%2Fapp%2Fpage.tsx%3A16%2C29

### Steps to Reproduce

1. Use `Portal` component in a React App

### Ark UI Version

1.1.0

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

cschroeter:
@aureldussauge 

Thanks for taking your time to create this issue. This is apparently a Codesandbox only problem since I could not reproduce this in a Next.js, Astro or component library setup but Codesandbox.

cschroeter:
@aureldussauge 

I've took a closer look and apparently TypeScript resolved the types to:

```ts
export declare const Portal: (props: PropsWithChildren<PortalProps>) => ReactNode;
```

Which is not correct. So I've added ean xplicit return type so the result looks like this:

```ts
export declare const Portal: (props: PropsWithChildren<PortalProps>) => ReactPortal[] | ReactNode | null | undefined;
```

I'm not sure if that helps but since its anyway just a Codesandbox thing I will close this issue. The "possible" fix for this will be availabe in the next release.

aureldussauge:
Thank you @cschroeter 

I have the problem in IntelliJ. I'll give it a try with the next release and keep you informed

PR #2155, opened 2024-02-09
fix(react): updates `Portal` to have backward compatibility

### LINKED ISSUE #2154: `Portal` return types are not compatible with some versions of `@types/react`.
### Description

Recently, after creating a simple Next.js project on my Windows machine and trying to use **Ark's** `Modal`, I got the following error:

```txt
Diagnostics:
1. 'Portal' cannot be used as a JSX component.
     Its return type 'ReactNode | ReactPortal[]' is not a valid JSX element.
       Type 'undefined' is not assignable to type 'Element | null'. [2786]
```

Replicating the same steps on another computer did not trigger the error.

Upon further investigation, I found the following issue: [DefinitelyTyped Pull Request #65135.](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65135)

It appears that **React v18** supports `ReactNode` as a return type for **Functional Components**. However, the `@types/react` package only updated its `JSXElementConstructor` definition to include `ReactNode` starting from version `18.2.8`.

### Before `18.2.8`

```ts
type JSXElementConstructor<P> =
  | ((props: P, deprecatedLegacyContext?: any) => ReactElement<any, any> | null)
  | (new (props: P) => Component<any, any>);
```

### Before After `18.2.8`

```ts
type JSXElementConstructor<P> =
  | ((props: P, deprecatedLegacyContext?: any) => ReactNode)
  | (new (props: P) => Component<any, any>);
```

All of this means that even with the type changes made in [#1787](https://github.com/chakra-ui/ark/issues/1787), users might still face compatibility issues if their project is using `@types/react` versions older than `18.2.8`. The discussion then is to decide if **Ark** will support "older" versions of the **React v18** types package.

There are a lot of approaches to fixing this issue, and if you decide to do so, I would really like to contribute by creating a PR.

No need for links

### Steps to Reproduce

1. Update `package.json`'s `@types/react` to a version older than `18.2.8`.
```json
  "devDependencies": {
    "@types/react": "18.2.7",
  }
```
2. Delete `node_modules`.
3. Install dependencies again.
4. Restart editor's `Lsp` if needed.

### Ark UI Version

^2.0.1

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

### Browser

_No response_

### Additional Information

_No response_

irwinarruda:
Fixes #2154 and #1787

There were a lot of ways this could be solved, most of them are by working around `typescript`. I decided to push the most type-safe solution since `Portal`'s return type now gets inferred as `JSX.Element`, so the code will work on all versions of `@types/react@18`. The only downside is adding a `Fragment` without actually needing it. 

I also added some tests for the `Portal` component since there weren't any :)

### DIFF
--- a/packages/frameworks/react/src/portal.test.tsx
+++ b/packages/frameworks/react/src/portal.test.tsx
@@ -1,0 +1,103 @@
+import { render, screen } from '@testing-library/react'
+import { useRef } from 'react'
+import { Portal } from './portal'
+
+const PortalWithContainerTest = (props: { disabled?: boolean }) => {
+  const container = useRef<HTMLDivElement>(null)
+  return (
+    <div>
+      <Portal container={container} disabled={props.disabled}>
+        <p>Anything must be visible</p>
+      </Portal>
+      <div data-testid="container" ref={container}></div>
+    </div>
+  )
+}
+
+describe('Portal', () => {
+  it('should render portal without element children', () => {
+    render(<Portal>Anything must be visible</Portal>)
+    expect(screen.getByText('Anything must be visible')).toBeVisible()
+  })
+
+  it('should render portal with element children', () => {
+    render(
+      <Portal>
+        <p>Anything must be visible</p>
+      </Portal>,
+    )
+    expect(screen.getByText('Anything must be visible')).toBeVisible()
+  })
+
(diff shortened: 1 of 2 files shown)


Omikorin:
Thanks!

PR #2271, opened 2024-03-08
fix: parse min and max date values props for date picker

tioluwani94:
This PR parses the min and max props for the date picker component.

### DIFF
--- a/packages/frameworks/react/src/date-picker/use-date-picker.ts
+++ b/packages/frameworks/react/src/date-picker/use-date-picker.ts
@@ -6,7 +6,10 @@ import type { Optional } from '../types'
 import { useEvent } from '../use-event'
 
 export interface UseDatePickerProps
-  extends Optional<Omit<datePicker.Context, 'value' | 'focusedValue' | 'open.controlled'>, 'id'> {
+  extends Optional<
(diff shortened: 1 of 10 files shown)


Omikorin:
Thank you @tioluwani94!

When you fix typings in the `date-picker-root.tsx` file and generate typedocs, I'll merge this PR

packages/frameworks/react/src/date-picker/use-date-picker.ts
```diff
@@ -19,6 +22,15 @@ export interface UseDatePickerProps
...
   value?: string[]
+  /**
+   * The minimum date for the date picker in the format yyyy-mm-dd
+   */
+  min?: string
+
+  /**
+   * The maximum date for the date picker in the format yyyy-mm-dd
+   */
+  max?: string
```
maastrich, line 33: @cschroeter @segunadebayo 
you may want to reference this change as breaking instead of fix as old calendarDate do not work anymore

----
btw that's too bad it was working quite well and zag is using calendar date
maybe we could use an union here ?

pierresisson: +1

PR #2364, opened 2024-04-05
docs: add install guides frameworks

carwack:
This PR splits the getting started page into install guides for frameworks like React, Solid and Vue.
I took the UX from PandaCSS and also recreated the framework cards for the introduction page.

Note: I have not yet double-checked the solid one in stackblitz. Done for Vue.


cschroeter:
Thanks for taking your time to create this. I've merged some ideas into the other existing PR. Very much appreciate <3

PR #2380, opened 2024-04-14
fix(docs): add missing theme switcher label

ludofischer:
Fix missing readable label on the theme switcher.


cschroeter:
@ludofischer 

Thanks buddy <3

PR #2423, opened 2024-04-23
WIP: fix: Vue histoire stories

carwack:
Currently going through all the Vue examples / stories in Histoire.

TODO:
- Select default example -> Type is String for select item but expected Object
- TagsInput BlurBehavior example -> strange behavior after trying to edit added tag
- Tooltip Controlled example -> strange behavior not showing tooltip


cschroeter:
@carwack 
Thanks buddy. i've addressed all those issues.

PR #2453, opened 2024-05-09
feat(docs): add new website

cschroeter:


iamdin:
Wondering why migrated the website from astro to next? @cschroeter

cschroeter:
@iamdin 

Have some ideas where I would like to fallback to next-auth and other toolings

PR #2484, opened 2024-05-25
docs: fix broken links in README

autologie:
This PR fixes broken links to components documentation in the README file.

I believe it's okay to fix it like this because the documentation seems to be identical for all supported frameworks and also it's easy to switch between frameworks on the website.


cschroeter:
@autologie 

Thanks buddy.

PR #2533, opened 2024-06-07
fix(svelte): fix context issues

Shyrro:
Main issue when using the docs example, is that the `$derived` rune, doesn't seem to retain reactivity correctly when paired with context. 

Instead, we create a state that we update ourselves. The `$state` rune reactivity seems to be retained when working with contexts.


packages/svelte/src/lib/components/avatar/avatar-root.svelte
```diff
@@ -1,21 +1,24 @@
...
-const avatar = useAvatar(props)
-AvatarProvider(avatar)
-</script>
+  AvatarProvider(avatar)
   
-<div {...avatar.getRootProps()}>
+</script>
+
+<div {...avatar.api.getRootProps()}>
+  <span {...avatar.api.getFallbackProps()}>PA</span>
```
anubra266, line 22:
```suggestion
```

packages/svelte/src/lib/components/avatar/use-avatar.svelte.ts
```diff
@@ -1,19 +1,16 @@
...
-  extends Optional<Omit<avatar.Context, 'dir' | 'getRootNode'>, 'id'> {}
-export interface UseAvatarReturn extends avatar.Api<PropTypes> {}
+  extends Optional<Omit<Context, 'dir' | 'getRootNode'>, 'id'> { }
+export interface UseAvatarReturn extends Api<PropTypes> { }
 
 export const useAvatar = (props: UseAvatarProps) => {
-  const [snapshot, send] = useMachine(avatar.machine({ id: '1' }))
-  const _api = avatar.connect(snapshot, send, normalizeProps)
-  const api = $derived(_api)
+  const stateMachine = useMachine(machine({ id: '1' }))
```
arekrgw, line 12: Why props are unused?

segunadebayo:
Thanks for working on this @Shyrro. Will fix it in Zag.js directly.

PR #2560, opened 2024-06-13
docs(qrcode): error correction

anubra266:
Add error correction section to qrcode docs

### DIFF
--- a/packages/react/src/components/qr-code/examples/with-overlay.tsx
+++ b/packages/react/src/components/qr-code/examples/with-overlay.tsx
@@ -2,7 +2,7 @@ import { QrCode } from '../..'
 
 export const Basic = () => {
   return (
-    <QrCode.Root value="http://ark-ui.com">
+    <QrCode.Root value="http://ark-ui.com" encoding={{ ecc: 'H' }}>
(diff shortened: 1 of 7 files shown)


packages/react/src/components/qr-code/examples/with-overlay.tsx
```diff
@@ -2,7 +2,7 @@ import { QrCode } from '../..'
 
 export const Basic = () => {
   return (
-    <QrCode.Root value="http://ark-ui.com" encoding={{ ecc: 'H' }}>
+    <QrCode.Root value="http://ark-ui.com">
```
segunadebayo, line 5:
```suggestion
    <QrCode.Root value="http://ark-ui.com" encoding={{ ecc: 'H' }}>
```

cschroeter:
@anubra266 

I combined the error correction and the overlay as you most likely need to do both as there is no real usecase for error correction if nothing is in front of the qr code.

segunadebayo:
@cschroeter That was my initial thought as well, but there's another scenario: when the link is very long.  That's another scenario you want to set the encoding.

Issue #2497, opened 2024-05-29
📌 Contribute to Ark UI: Components Ready for Implementation


cschroeter:
## Welcome Contributors!

Thank you for your interest in contributing to Ark UI! Below is a list of components that are available for implementation. If you're looking to make a meaningful contribution, this is a great place to start.

### Available Components

| Component      | Link                                                                                          | Status       | PR/Contributor |
| -------------- | --------------------------------------------------------------------------------------------- | ------------ | -------------- |
| Floating Panel | [Floating Panel](https://github.com/chakra-ui/zag/tree/main/packages/machines/floating-panel) | In Progress  | @mverissimo    |
| Signature Pad  | [Signature Pad](https://github.com/chakra-ui/zag/tree/main/packages/machines/signature-pad)   | ✅ Completed | @jeferson-sb   |
| Steps          | [Steps](https://github.com/chakra-ui/zag/tree/main/packages/machines/steps)                   | ✅ Completed | @segunadebayo  |
| Time Picker    | [Time Picker](https://github.com/chakra-ui/zag/tree/main/packages/machines/time-picker)       | ✅ Completed | @colinlienard  |
| Timer          | [Timer](https://github.com/chakra-ui/zag/tree/main/packages/machines/timer)                   | In Progress  | @jeferson-sb   |

### How to Get Started

1. **Check the Component's Status**: Ensure the component you want to work on is marked as "Not Started."
2. **Comment Below**: Express your interest in implementing the component by leaving a comment.
3. **Fork the Repository**: Make a fork of the Ark UI repository to start working on your component.
4. **Implement the Component**: Look up existing components in the repository and follow their blueprint to maintain consistency.
5. **Submit a Pull Request**: Once done, submit a pull request for review.

**Happy coding! 🚀**

jeferson-sb:
I would like to take a shot on the Signature Pad

jeferson-sb:
hey can I work on the Timer as well?

colinlienard:
Hey, I would like to continue my work on the time picker here

segunadebayo:
Thanks to everyone who contributed so far. We'll focus on getting the draft PRs in the next weeks.

PR #2590, opened 2024-06-21
feat(*): implement timer

jeferson-sb:
Implement Zag's Timer in React
as part of #2497 

There's one issue still to be solved related to the action trigger, it seems that click handler from `timer.getActionTriggerProps()` does not modify the machine 🤔 and I couldn't figure out yet.

Demo

https://github.com/chakra-ui/ark/assets/30840709/abe8d26c-751d-4319-a917-9d3d9d476cfb


jeferson-sb:
It seems the issue from the getActionTriggerProps props might come from the actions on Zag.js

https://github.com/chakra-ui/zag/blob/70e6f64acc997490ef5c035f0995c6e8c9fd59d9/packages/machines/timer/src/timer.connect.ts#L75-L85

I wonder if it's expecting `PAUSE`, `START`, `RESUME` all uppercase instead of lowercase.

cschroeter:
@jeferson-sb

I've updated your PR. You were right. A fix in zag-js will be included in the next release.

PR #2651, opened 2024-07-09
feat(*): add floating panel

mverissimo:
Add the react implementation of @zag-js/floating-panel as part of https://github.com/chakra-ui/ark/issues/2497

@cschroeter I have some issues, could you help me?

1. An unexpected behavior happens when the window is resized:

https://github.com/chakra-ui/ark/assets/371844/6591bc3d-34f3-4658-8167-bbdb74ac040d

2. Following the blueprint, the API(PublicContext) doesn't export the type `open.controlled` when implementing the' controlled' mode.

3. In the anatomy, a `part` called `dock ` is exported, but this look is not used internally


segunadebayo:
Thanks for starting this @mverissimo. I've gone ahead and completed it for the other frameworks.

PR #2799, opened 2024-08-25
fix(vue): presence of componets with `Positioner`

### LINKED ISSUE #2739: Some components can't have animation on exit
### Description

When I use a dialog component for example, and add animation classes for enter and exit, only enter animation works. It's like the exit animation is being terminated by the attribute that Ark adds on multiple elements.

Screenshot 2024-08-05 at 10 14 43

https://stackblitz.com/edit/vitejs-vite-xmvqqu

### Steps to Reproduce

1. Click the button
2. Dialog shows up with animation on enter
3. Close the dialog and there's no exit animation

Tailwind classes are in place and animations are inside of tailwind config.

### Ark UI Version

3.7.2

### Framework

- [ ] React
- [ ] Solid
- [X] Vue

### Browser

All

### Additional Information

_No response_

YaredFall:
Resolves #2739.

As stated in the issue, any component with `Positioner` part has an issue with an exit animation, because `Presence` is used on the `Positioner` instead of a `Content`.

This PR fixes all of the related vue components. 

Code changes are made taking React implementation as an example and briefly tested in histoire stories.

### DIFF
--- a/packages/vue/src/components/color-picker/color-picker-content.vue
+++ b/packages/vue/src/components/color-picker/color-picker-content.vue
@@ -24,7 +24,7 @@ const mergedProps = computed(() => mergeProps(colorPicker.value.getContentProps(
 </script>
 
 <template>
-  <ark.div v-bind="mergedProps">
+  <ark.div v-if="!presence.unmounted" v-bind="mergedProps" :as-child="asChild">
     <slot />
   </ark.div>
 </template>
--- a/packages/vue/src/components/color-picker/color-picker-content.vue
+++ b/packages/vue/src/components/color-picker/color-picker-content.vue
@@ -1,7 +1,8 @@
 <script lang="ts">
-import type { HTMLAttributes } from 'vue'
+import { mergeProps } from '@zag-js/vue'
+import { type HTMLAttributes, computed } from 'vue'
 import type { PolymorphicProps } from '../factory'
-import type { PresenceProps } from '../presence'
+import { type PresenceProps, usePresenceContext } from '../presence'
 
 export interface ColorPickerContentBaseProps extends PresenceProps, PolymorphicProps {}
 export interface ColorPickerContentProps
@@ -18,10 +19,12 @@ import { useColorPickerContext } from './use-color-picker-context'
 
 defineProps<ColorPickerContentProps>()
 const colorPicker = useColorPickerContext()
+const presence = usePresenceContext()
+const mergedProps = computed(() => mergeProps(colorPicker.value.getContentProps(), presence.value.presenceProps))
 </script>
 
 <template>
-  <ark.div v-bind="colorPicker.getContentProps()">
+  <ark.div v-bind="mergedProps">
     <slot />
   </ark.div>
 </template>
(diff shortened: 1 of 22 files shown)


Omikorin:
Awesome @YaredFall!

I approved a workflow. The dialog and tooltip tests failed, could you look into it? I'll rerun them to be sure.

I'll review this PR today

YaredFall:
As for `Tooltip` I indeed forgot to put `v-if` directive. Fixed it now. Test seems to be failing nonetheless, although I manually tested it inside histoire and lazy mounting is now working.
As for `Dialog`, idk... It's working as expected inside histoire but.
I'm not familiar with Testing Library, so maybe I miss something 😕

Omikorin:
> As for `Tooltip` I indeed forgot to put `v-if` directive. Fixed it now. Test seems to be failing nonetheless, although I manually tested it inside histoire and lazy mounting is now working. As for `Dialog`, idk... It's working as expected inside histoire but fails test. I'm not familiar with Testing Library, so maybe I miss something 😕

Thank you for working on that.

Your code is working well. The issue with tests is that they are not waiting enough time for a specified behavior. It's a visible case with the slowly exiting elements, like the exit animation. Lazy mounting is a problem also, because we need time to remove an element from the DOM.

Please take a look at the [hover card test](https://github.com/chakra-ui/ark/blob/3ed7ad64440f900a1a72e7f9cbba8c78ce0e6196/packages/vue/src/components/hover-card/tests/hover-card.test.ts#L28).

Sometimes it's not the problem of a code, but a random timeout on GitHub Actions (the free tier is very slow:/ ). If nothing will work, I'll take a look myself!

YaredFall:
Thank you for your feedback and insights!

I've adjusted the tests, simply by wrapping expect statements with `waitFor` function. For some reason the `Dialog` test also needed an `await` before `screen.findByText('Dialog Title')` as it was returning the whole document without it 😕

Now tests are passing, but I feel a bit uncretain about it

Omikorin:
> I've adjusted the tests, simply by wrapping expect statements with `waitFor` function. For some reason the `Dialog` test also needed an `await` before `screen.findByText('Dialog Title')` as it was returning the whole document without it 😕

That's right. Sometimes there are race conditions, because it's not only a test, but we need to make sure the test gives us some time to render everything before continuing its workflow. You need to remember tests are slower on the CI/CD runner than on your computer.

We run tests in an async manner but at the same time there are tests that their code needs to be run sequentially. I found that case mostly connected to a component's complexity and the time of its reaction. Look at the `DatePicker` component. This component is so huge that tests were failing simply of timeout or being too fast for an expected behavior (i.e. lazy mounting - removing a lot of DOM elements).

If you are interested you can check the difference between React and Vue tests. I enchanced React tests because its Testing Library wrapper allows to `describe` tests that a rendered component is cleaned after all cases ended (`afterAll` vs `afterEach`). There is the `retry: 2` option set in the global configuration to silence false-positives too.

For example, `getParts` should render a component once and check everything. We could do it by hard-coding each Part, but then we have a problem with a clear message what failed).

To put it simply, testing can be awful with keeping everything nice and fast :sweat_smile: 

I think everything is working, so I'll double check manually this PR after work!

Omikorin:
Thank you @YaredFall, nice job! I appreciate this
Every commit is clean and you write code straightforward without any critical errors :)

Issue #2665, opened 2024-07-15
Select component do not sync `selected` state to native select element


Reverier-Xu:
### Description

hi, i'm building a custom select component using @ark-ui/solid and [modular forms](https://modularforms.dev/solid/guides/special-inputs), i found that the select component does not sync selected option to native select element inside `Select.HiddenSelect`.

### Link to Reproduction (or Detailed Explanation)

https://codesandbox.io/p/sandbox/solid-select-8njd3j?file=%2Findex.tsx%3A50%2C1

### Steps to Reproduce

1. open the ark-ui select box and select something;
2. open DevTools and navigate to the native select;
3. options are all un-selected here:

![image](https://github.com/user-attachments/assets/fe4a3c25-318d-4539-b14d-16a61a6eef20)

### Ark UI Version

3.5.0

### Framework

- [ ] React
- [X] Solid
- [ ] Vue

segunadebayo:
This might not reflect directly in the DOM, but if you log the `selectEl.value` or use it within a form, you should see the selected value.

![CleanShot 2024-07-15 at 18 26 39@2x](https://github.com/user-attachments/assets/e4807ced-7284-4edd-b406-d76cadf9b8ac)

I believe this has something to do with the integration with modular-forms. If you have more details, feel free to create a discussion about this.

YogiLiu:
@segunadebayo  Got the similar problem, but in my case, all the value of options are empty.

![image](https://github.com/user-attachments/assets/c4b73e9f-f71f-4799-bf52-f8c7459cd036)

## Steps to Reproduce

1. Create a typescript solid project using `pnpm create vite@latest` .
2. Execute `pnpm i` in project root to install dependencies. 
3. Execute `pnpm add @ark-ui/solid`.
4. Remove all the code in the `index.css`.
5. Paste the following code to `App.tsx`.
6. Execute `pnpm run dev` to start the dev server.

This is my minimal reproducible code:

```tsx
import { Index, Portal } from "solid-js/web";
import { Select } from "@ark-ui/solid";

function App() {
  const items = ["React", "Solid", "Vue"];
  return (
    <form
      onChange={(e) => {
        const fd = new FormData(e.currentTarget);
        console.log("framework:", fd.get("framework"));
      }}
    >
      {/* Copy from https://ark-ui.com/solid/docs/components/select#examples */}
      <Select.Root name={"framework"} items={items}>
        <Select.Label>Framework</Select.Label>
        <Select.Control>
[...]

PR #2814, opened 2024-08-26
fix(solid): fix option value and disabled in hidden select

YogiLiu:
The item type of Index component is `() => T`, we should call it to access the value.

Related: #2665


YogiLiu:
Are there any plans to release a new version to fix this problem? @segunadebayo

PR #2881, opened 2024-09-13
[Vue] Expose `v-model` for `Field` form fields

TylerAPfledderer:
Allows usage of the `v-model` from `FieldInput`, `FieldSelect`, and `FieldTextarea`.

The approach to set `v-model` at these components instead of at `FieldRoot` breaks the pattern with other Ark components, as the value type is variable depending on the component used, and therefore need to retain the type. This is consistent with how `value` is handled in the React version.


packages/vue/src/components/field/field-input.vue
```diff
@@ -8,17 +8,28 @@ export interface FieldInputProps
...
+  modelValue?: InputHTMLAttributes['value']
+}
 </script>
 
 <script setup lang="ts">
 import { ark } from '../factory'
 import { useFieldContext } from './use-field-context'
 
-defineProps<FieldInputProps>()
+const props = defineProps<FieldInputProps>()
```
carwack, line 20: It shouldn't be needed to add the `const props =` since it's only used in the `<template>` and there it knows the props and you can call them through `modelValue` instead of `props.modelValue`.

packages/vue/src/components/field/field-input.vue
```diff
@@ -8,17 +8,28 @@ export interface FieldInputProps
     /**
      * @vue-ignore
      */
-    InputHTMLAttributes {}
+    Omit<InputHTMLAttributes, 'value'> {
+  modelValue?: InputHTMLAttributes['value']
```
carwack, line 12: Have you looked into the helper `defineModel()` https://vuejs.org/guide/components/v-model.html?

I do think this removes the need to add the prop and emit in this case. But I'm currently not really aware if the emit for example is needed specifically somewhere.

packages/vue/src/components/field/field-select.vue
```diff
@@ -8,17 +8,28 @@ export interface FieldSelectProps
...
+const emit = defineEmits(['update:modelValue'])
 </script>
 
 <template>
-  <ark.select v-bind="field.getSelectProps()" :as-child="asChild"><slot /></ark.select>
+  <ark.select
+    v-bind="field.getSelectProps()"
+    :value="props.modelValue"
+    @change="(event) => emit('update:modelValue', (event.target as HTMLSelectElement).value)"
+    :as-child="asChild"
```
carwack, line 31: If the v-bind name and the value name is the same, it's now (since v3.4 of Vue) not needed to type them both. It can just be `:as-child`.

```
<!-- same-name shorthand (3.4+), expands to :src="src" -->
<img :src />
```

packages/vue/src/components/field/tests/field.test.vue
```diff
@@ -1,11 +1,12 @@
 <script setup lang="ts">
 import { Field } from '../..'
+const model = defineModel()
```
carwack, line 3: I'm not 100% sure this is the correct way to use the `defineModel()` helper. Please ignore if I'm wrong here 😄

carwack (commented):
Just a few improvements. Thanks for the PR! 💪

packages/vue/src/components/field/field-input.vue
TylerAPfledderer: I tried it and it wasn't yielding the expected results. Figured to take the old approach for now and reevaluate the issue later.

packages/vue/src/components/field/tests/field.test.vue
TylerAPfledderer: For this purpose, it's to emit the model up a second level for this test file

packages/vue/src/components/field/field-select.vue
TylerAPfledderer: That was already there. But I can go ahead and update it here since it's small 😁

packages/vue/src/components/field/field-input.vue
TylerAPfledderer: I tried applying it again, and the value just does not want to emit. Maybe need to take a look at the factory

TylerAPfledderer: Yep, something is happening in the factory to prevent `defineModel` from functioning at the component level

carwack: Alright, let's keep it this way for now and we can improve it in the future

PR #2970, opened 2024-10-09
fix(react): avoid unnecessary re-render in useIsServer

### LINKED ISSUE #2965: Dialog with defaultOpen closes itself immediatly after first render
### Description

When I set `defaultOpen` prop of the Dialog to `true` the Dialog is usually shown only for a fraction of a second and then it's closed. Rarely the Dialog stays open, but then it exhibits another issue, that clicking inside the Dialog.Content closes the Dialog. Even more rarely the Dialog stays open and behaves as expected - maybe like 1 in 30 renders.

I've encountered similar issue when setting the prop `open` directly to `true`.

https://codesandbox.io/p/devbox/silent-thunder-5fdk7q

### Steps to Reproduce

1. Load the page
2. See the modal, containing text "TEST", is closed (only shown for a fraction of second)

### Ark UI Version

4.1.0

### Framework

- [X] React
- [ ] Solid
- [ ] Vue

### Browser

Firefox, Chromium - latest versions

### Additional Information

_No response_

hrsh7th:
The current implementation of `useIsServer` causes the render to happen twice.

This PR reduces re-render by using `useSyncExternalStore` way.

Closes #2965


packages/react/src/components/portal/portal.tsx
```diff
@@ -10,13 +10,20 @@ export interface PortalProps {
...
+  const [container, setContainer] = useState(props.container?.current)
+  const { children, disabled } = props
   const isServer = useIsServer()
   const { getRootNode } = useEnvironmentContext()
 
+  useEffect(() => {
+    setContainer(() => {
+      return props.container?.current
+    })
+  }, [props.container?.current])
```
hrsh7th, line 22: The reason the previous implementation passed the tests is because `useIsServer` causes a re-render.

This useEffect emulates the old behavior.

cschroeter:
@hrsh7th 

Thanks your for your PR. Is there any chance to provide some code so we can understand the difference between those two approaches?

From what we can tell, is that the `useState` is still updated using an effect so really would love to see some "proof" that it works as intended.

Cheers

hrsh7th:
Thank you for your comment.

The following test passed on my end. So the new implementation seems to reduce the number of renderings in the case where `container` is not specified.

However, the test code is ugly and I am not sure if it will be acceptable.
Should I also push the following test code?

```
  it('should render twice if container was specified', async () => {
    const RenderCount = () => {
      RenderCount.count++
      return <div>{RenderCount.count}</div>
    }
    RenderCount.count = 0

    const Test = () => {
      const ref = useRef<HTMLDivElement>(null)
      return (
        <div>
          <Portal container={ref}>
            <RenderCount />
          </Portal>
          <div ref={ref} />
        </div>
      )
    }
    const view = render(<Test />)
    await waitFor(() => expect(screen.getByText('2')).toBeInTheDocument())
    expect(view.baseElement).toMatchInlineSnapshot(`
      <body>
        <div>
          <div>
            <div>
              <div>
                2
              </div>
            </div>
          </div>
        </div>
      </body>
    `)
  })

  it('should render once if container was not specified', async () => {
    const RenderCount = () => {
      RenderCount.count++
      return <div>{RenderCount.count}</div>
    }
    RenderCount.count = 0
    const Test = () => {
      return (
        <div>
          <Portal>
            <RenderCount />
          </Portal>
        </div>
      )
    }
    const view = render(<Test />)
    await waitFor(() => expect(screen.getByText('2')).toBeInTheDocument(), {
      timeout: 500,
    }).catch(() => {})
    expect(view.baseElement).toMatchInlineSnapshot(`
      <body>
        <div>
          <div />
        </div>
        <div>
          1
        </div>
      </body>
    `)
  })
```

PR #3251, opened 2025-01-25
refactor(react): compiled zag

segunadebayo:
Improve initial render performance by inlining the state machine logic using our internal Zag.js compiler. This will reduce the number of re-renders and improve the overall performance of the components.


segunadebayo:
Designed a better version of Zag, which will be our v1. We'll use that instead.

Issue #3286, opened 2025-02-05
Field HelperText does not receive data-disabled attribute


veloware:
### Description

When I set `disabled` on the `Field.Root` I expect `data-disabled` to be added to `Field.HelperText` so that I can style it accordingly (Similar to `Field.Label`)

### Link to Reproduction (or Detailed Explanation)

Setting `data-disabled` on the helper text will allow for styling of the helper text similar to the label

### Steps to Reproduce

1. Render `Field.Root` with `disabled` prop
2. Render a nested `Field.HelperText` with styles attached that checks for `[data-disabled]` (or `_disabled`) with Chakra
3. Notice the style does not take effect.

```
<Field.Root disabled>
   <Field.HelperText>Helper Text</Field.HelperText>
</Field.Root>
```

### Ark UI Version

4.9.1

### Framework

- [x] React
- [ ] Solid
- [ ] Vue

### Browser

All

### Additional Information

`useField.ts` could add - 

```
'data-disabled': dataAttr(disabled),
```

to the `getHelperTextProps` function.

segunadebayo:
Feel free to open a PR to add this to all framework

veloware:
will do, though, I'm getting this when running `npm i`, any suggestions?

<img width="611" alt="Image" src="https://github.com/user-attachments/assets/1eddd592-eb39-494c-b9fa-8c88a207ee31" />

carwack:
@veloware Could you use `bun` instead of `npm`? The Ark monorepo is set up with `bun`. I assume this would also solve your npm resolve error.

PR #3304, opened 2025-02-15
Fix: Field Helper Text missing data-disabled attribute when field is disabled

carwack:
Added `data-disabled` attribute to `getHelperTextProps()` for the Field component (React, Solid and Vue).
Added disabled example for the Field component (React, Solid and Vue).
Added tests to check on the `data-disabled` attribute when the Field is disabled  (React, Solid and Vue).

Fixes: #3286


cschroeter:
@carwack awesome mate. thanks for your contribution

PR #3307, opened 2025-02-16
feat(field): add aria-describedby to Field component

### LINKED ISSUE #3301: FieldInput does not receive aria-describedby when FieldHelper is used.
### Description

FieldInput should automatically get `aria-describedby` when a FieldHelper is present within the FieldRoot, but it doesn't.

https://ark-ui.com/react/docs/components/field

### Steps to Reproduce

Inspect the Field Example above.

### Ark UI Version

4.9.1

### Framework

- [x] React
- [x] Solid
- [x] Vue

### Browser

Chrome, Firefox, Safari

### Additional Information

_No response_

jeferson-sb:
## Description

Add aria-describedby based on the Field helper text
closes #3301


packages/react/src/components/field/use-field.ts
```diff
@@ -137,8 +137,9 @@ export const useField = (props: UseFieldProps) => {
       ({
         ...getControlProps(),
         ...parts.input.attrs,
+        'aria-describedby': helperTextId,
```
carwack, line 140: I do believe that `aria-describedby` is already part of the `getControlProps()`, see line 122 of this file.
The issue lies more with the `labelsIds` array not being filled correctly.

carwack (commented):
Thanks for contributing! 🙏

I do left my remarks on this issue. 
Also, is it possible if you find the solution to make it happen for the other frameworks too (Vue and Solid)?

Thanks!

jeferson-sb:
@carwack Is there an option to run storybook for Solid or Vue, I couldn't find one

jeferson-sb:
@carwack Can you also check why some of these files got formattted the way they are?

carwack:
Hi @jeferson-sb,
Thanks for the work!

> @carwack Is there an option to run storybook for Solid or Vue, I couldn't find one

You can use the following commands
Solid: `bun solid storybook` -> we noticed there is a bug on the solid storybook package which might break your build (no solution so far so if that happens, we have to skip it)/
Vue: `bun vue storybook` -> currently using Histoire for the stories.

> @carwack Can you also check why some of these files got formattted the way they are?
I noticed this as well, I wonder if this is because of new or different prettier rules. I will have to look into this.

carwack:
> @carwack Can you also check why some of these files got formattted the way they are?

@cschroeter I noticed the formatting happens in the `lefthook` step on committing the file. The `format` step does what I expect as stated in the `.prettierrc` config file. 
But the `lint` step, which uses Biome.js makes the changes to add the colins, the double quotes and spacing back. 
I have not enough knowledge to know exactly where it comes from, was there an update in the `recommended` rules for example of Biome?

carwack:
@jeferson-sb for the formatting. I believe I found the issue. I got a fix here: https://github.com/chakra-ui/ark/pull/3316

Maybe you can test it locally yourself by adding this to the `biome.json` file:
```json
"formatter": {
  "enabled": false
}
```
I expect that then, when you commit code changes, prettier formats it and biome only does the lint checking, and not also format it (which has no setup for formatting)

segunadebayo:
Thanks for this PR. I've handled this in the zag v1 PR which will be merged pretty soon.

PR #3311, opened 2025-02-18
fix(menu): close menu on Tab key press

pablopupo:
When pressing Tab in the menu, the menu remained open, which is not expected behavior for keyboard navigation according to [WAI ARIA Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/patterns/) for [Menu and Menubar Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/). This was an issue that was reported in [Issue 9608 of chakra-ui/chakra-ui](https://github.com/chakra-ui/chakra-ui/issues/9608) but needed to be fixed upstream. 

I fixed this by  adding an onKeyDown handler in menu-content.tsx that calls menu.setOpen(false) when event.key === 'Tab'


cschroeter:
@pablopupo 

Thanks for taking your time to create the PR. While your proposal sounds good at first, menus can also contain links so closing a menu on tab is not always the best approach. I would recommend to keep this logic in your menu component and everything should be fine.

```tsx
const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
  if (event.key === "Tab") {
  // Instead of menu?.close?.(), use setOpen(false)
  menu?.setOpen?.(false);
}
  // Call any original onKeyDown prop
  onKeyDown?.(event);
};
```

PR #3315, opened 2025-02-21
fix(react): add focus-trap barrel export

### LINKED ISSUE #3314: FocusTrap component not available via main import in React
### Description

When I wanted to use the new FocusTrap component I expected to be able to import it through "@ark-ui/react" but it was not available

https://codesandbox.io/p/devbox/quizzical-monad-7rzsgk

### Steps to Reproduce

1. Try to `import { FocusTrap } from "@ark-ui/react";`

### Ark UI Version

4.9.1

### Framework

- [x] React
- [ ] Solid
- [ ] Vue

### Browser

_No response_

### Additional Information

It is available for Vue and Solid, just not React.

ChrisRu:
closes #3314 

Looked at other imports too but they all seemed to be in place. And Solid and Vue already have the barrel export, it was just React that's missing.

Could consider adding a verification step for this to some script to make sure it does not happen again? Looked into `scripts/src/check-exports.ts` briefly, but I did not understand fully and I'm not sure that's the place for it.


carwack:
Thanks for this!

I personally think we can merge this and add the proposal of adding a verification step on our to-do list?

ChrisRu:
Cool, sounds good to me!

PR #3317, opened 2025-02-21
fix: render option labels inside hidden select

Jonas-C:
This was picked up by an accessibility scanner we use. Zag already seems to render the labels in their examples, so it should probably be done here as well.

Excuse the formatting changes, the change itself is actually quite small.


cschroeter:
@Jonas-C  thx mate

PR #3396, opened 2025-03-18
Feat/add vue storybook

carwack:
Add Storybook v8.6.7 in favor of Histoire for the Vue package.

As reply on: https://github.com/chakra-ui/ark/pull/3382#issuecomment-2717793227

Works with one workaround in `.storybook/main.ts`
```
 // TODO: fix this when vue-docgen accepts types from other packages in components
   // This is a workaround until vue-docgen-plugin is fixed since we also dont use autodocs
   // This removes the plugin vue-docgen-plugin so it wont load and throws an error on the use of an type import like
   // import { Avatar, type AvatarRootEmits, type AvatarRootProps } from '@ark-ui/vue/avatar'
   // using options: { docgen: "vue-component-meta",} throws other errors with node_modules/@zag-js/avatar/dist/index.mjs.vue' 
   // could not be found. This could be a fix for the vue-docgen-plugin error but it is not working for now
   viteFinal(config) {
     const vueDocgenIndex = config.plugins?.findIndex((plugin) => 
       plugin && typeof plugin === 'object' && 'name' in plugin && plugin.name === 'storybook:vue-docgen-plugin'
     ) ?? -1
     if (vueDocgenIndex !== -1) config.plugins?.splice(vueDocgenIndex, 1)
     return config
   }
   ```

### CODE REVIEW

File: packages/vue/src/components/accordion/accordion.stories.ts [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+  render: () => ({
+    components: { DisabledExample },
+    template: '<DisabledExample />',
+  }),
+}
+
+export const Closed = {
+  render: () => ({
+    components: { AccordionExample },
+    template: `<AccordionExample multiple :items="['React', 'Solid', 'Vue']" />`,
```
cschroeter, line 83: @carwack 

I don't know but having the template as a `string` feels not great from a DX perspective, as errors can easily happen. Is there an alternative to this?

carwack: I just saw this review, thanks for creating it and sorry for the long wait on a response. 
I don't know yet, but I will take a look into it asap and see if there is an alternative!

carwack: I have been looking into it. Mainly checked the docs of Storybook. For Vue it seems that we either have to set the most complex component of the examples as the reference in the meta and then be able to just set the args for that story. The other stories still need to have the template set.
```
const meta = {
  title: 'Components / Avatar',
  component: ClosedExample, // Reference component for Storybook docs
  argTypes: {
    name: { control: 'text', description: 'The user name (used for initials fallback)' },
    src: { control: 'text', description: 'URL to the avatar image' },
  },
} satisfies Meta<typeof ClosedExample>

export default meta
type Story = StoryObj<AvatarBaseProps>
...
export const Basic: Story = {
  render: () => ({
    components: { BasicExample },
    template: '<BasicExample />',
  }),
  args: {},
}

export const Closed: Story = {
  args: {
    name: "Christian",
    src: "https://avatars.githubusercontent.com/u/1846056?v=4",
  },
}
```

Alternative is using the args like this:
```
export const Closed = {
  render: (args: AvatarProps) => ({
    components: { ClosedExample },
    setup() {
      return {
        args,
        onStatusChange: action('status changed'),
      };
    },
    template: '<ClosedExample v-bind="args" @status-change="onStatusChange" />',
  }),
  args: {
    name: "Christian",
    src: "https://avatars.githubusercontent.com/u/1846056?v=4",
  } satisfies AvatarProps,
}
```
https://storybook.js.org/docs/api/csf

Altho it's mostly needed like that when we have `closed` component examples. Other examples are mostly just the template like:
```
export const Basic = {
  render: () => ({
    components: { BasicExample },
    template: '<BasicExample />',
  }),
}
```

I feel like the 2nd option is the best here. What do you think of it?

carwack: I will look into implementing the 2nd option in this PR asap. @cschroeter 

```ts
export const Closed = {
  render: (args: AvatarProps) => ({
    components: { ClosedExample },
    setup() {
      return {
        args,
        onStatusChange: action('status changed'),
      };
    },
    template: '<ClosedExample v-bind="args" @status-change="onStatusChange" />',
  }),
  args: {
    name: "Christian",
    src: "https://avatars.githubusercontent.com/u/1846056?v=4",
  } satisfies AvatarProps,
}
```

```ts
export const Basic = {
  render: () => ({
    components: { BasicExample },
    template: '<BasicExample />',
  }),
}
```

segunadebayo:
Thanks for this PR. I've been in doubt whether to accept this PR or not, given the amount of work that has gone into it.

I'm saying this because I know you can handle this. To prevent the sunk cost fallacy, I suggest we park this PR for now. The current Vue setup is sufficient for most use cases. I no longer see any added benefit in migrating to Storybook.

Sincerely apologize for leading us down this path.

PR #3398, opened 2025-03-19
Fix Button Loading State Logic (react)

hkwid:
## Description

This PR fixes a logic issue in the Button component where loadingText would always be displayed when provided, regardless of the loading state. The correct behavior is to only show loadingText when the button is in a loading state.

## Changes

Updated the conditional rendering logic in the Button component
The component now correctly:

Shows loadingText if loading=true and loadingText exists
Shows spinner with invisible children if loading=true and no loadingText
Shows children if loading=false, regardless of loadingText value

### DIFF
--- a/website/src/components/ui/button.tsx
+++ b/website/src/components/ui/button.tsx
@@ -17,13 +17,15 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
 
   return (
     <StyledButton disabled={trulyDisabled} ref={ref} {...rest}>
-      {loading && !loadingText ? (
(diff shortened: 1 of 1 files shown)


carwack:
Thanks for this fix!

This looks good to me, what do you guys think @segunadebayo @cschroeter? (not a react expert myself)

segunadebayo:
Thanks for opening this. It should be fixed in Park UI not here.

Please consider opening this PR there.

hkwid:
@segunadebayo
Thanks. I opened the PR to park-ui but no response for more than 1 week...

PR #3417, opened 2025-03-27
fix(solidjs): keep textarea component reactive

### LINKED ISSUE #3416: [solidjs] Textarea does not properly react to controlled values
### Description

While using Ark `Field.Textarea` with `tanstack/solid-form` I stumbled upon some weird behaviour. I tried making a minimal repro and it can be boiled down to just having a value field controlled by a signal.

The best case to notice this is clearing the textarea value, and with a vanilla textarea next to it you can see it fairly clearly:

Controlled values in `Field.Textarea` get ignored outside of its' initial value, thus outside updates do not 

https://gist.github.com/jeroenptrs/81191a2ab3a06ac020224a6c4db4ced4

### Steps to Reproduce

1. Using the SolidStart example in https://ark-ui.com/react/docs/overview/getting-started, change routes/index.tsx to the linked gist (https://gist.github.com/jeroenptrs/81191a2ab3a06ac020224a6c4db4ced4)
2. Type some values in the Field.Textarea component (you should see it duplicated in the vanilla textarea)
3. Press the clear button
4. The value should be gone in the vanilla textarea, but still present in Field.Textarea

### Ark UI Version

5.1.0

### Framework

- [ ] React
- [x] Solid
- [ ] Vue

### Browser

Chrome and Safari at the least

### Additional Information

_No response_

jeroenptrs:
Fixes #3416

### DIFF
--- a/packages/solid/src/components/field/field-textarea.tsx
+++ b/packages/solid/src/components/field/field-textarea.tsx
@@ -1,6 +1,6 @@
 import { autoresizeTextarea } from '@zag-js/auto-resize'
 import { mergeProps } from '@zag-js/solid'
-import { createEffect } from 'solid-js'
+import { createEffect, splitProps } from 'solid-js'
 import { type HTMLProps, type PolymorphicProps, ark } from '../factory'
 import { useFieldContext } from './use-field-context'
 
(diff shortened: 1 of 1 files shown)


segunadebayo:
Thanks @jeroenptrs 

Pushed a fix for this already

PR #3419, opened 2025-03-29
fix(template-nuxt): update quickstart doc and add TS as dev dependency

Us3r-gitHub:
### **Summary**
This PR includes two small updates:
1. Fixed the `href` property in the Nuxt Framework quickstart documentation.
2. Added TypeScript as a dev dependency to the Nuxt template.

### **Changes**
- **Documentation**: Corrected the `href` in the Nuxt quickstart guide.
- **Template**: Added TypeScript as a dev dependency.

### DIFF
--- a/templates/nuxt/package.json
+++ b/templates/nuxt/package.json
@@ -14,5 +14,8 @@
     "nuxt": "3.16.1",
     "vue": "3.5.13",
     "vue-router": "4.5.0"
+  },
+  "devDependencies": {
+    "typescript": "5.8.2"
   }
 }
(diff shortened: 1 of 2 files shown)


Us3r-gitHub:
I noticed that the previous commit [fix(website): resolve to github instead] (ab61279) aimed to update the URL to GitHub, but the current code still uses the StackBlitz URL.

Should we keep the StackBlitz URL, or update it to the GitHub URL?

segunadebayo:
Let’s keep the Stackblitz URL for now. We want to use the GitHub for local experimentation.

PR #3454, opened 2025-04-22
feat(website): use ark ui design system for docs

cschroeter:


segunadebayo:
Closing this. Not worth the effort (as we discussed), given the number of frameworks involved.

We can revisit this in the future. I smell there are alternative strategies we can explore.

PR #3562, opened 2025-07-28
fix: call `React.isValidElement()` before `Children.only()`

### LINKED ISSUE #3561: Check whether `children` is a valid element with `React.isValidElement()` before calling `Children.only()`
### Description

We suspect a same bug of https://github.com/chakra-ui/chakra-ui/issues/10191 exists on Ark UI too since https://github.com/chakra-ui/ark/blob/0a5a1548570e0f2b396eab2e4b68951b747cd6fe/packages/react/src/components/factory.ts#L51-L61 calls `Children.only()` without `React.isValidElement()`.
However, I failed to create a reproduction. As I wrote on https://github.com/chakra-ui/chakra-ui/issues/10191, if we change something a bit, the error easily disappears.

I failed to create a reproduction for Ark UI, but, created for Chakra v3 https://github.com/exKAZUu/chakra-v3-issue

### Steps to Reproduce

N/A

### Ark UI Version

N/A

### Framework

- [x] React
- [ ] Solid
- [ ] Vue
- [ ] Svelte

### Browser

Google Chrome 138.0.7204.169

### Additional Information

_No response_

exKAZUu:
Closes #3561

We've tested this change by modifying `node_modules`. cf. https://github.com/chakra-ui/ark/issues/3561#issuecomment-3126863671

### DIFF
--- a/packages/react/src/components/factory.ts
+++ b/packages/react/src/components/factory.ts
@@ -53,12 +53,12 @@ const withAsChild = (Component: React.ElementType) => {
       if (!asChild) {
         return createElement(Component, { ...restProps, ref }, children)
       }
-
-      const onlyChild: React.ReactNode = Children.only(children)
-
-      if (!isValidElement<Record<string, unknown>>(onlyChild)) {
+      
+      if (!isValidElement<Record<string, unknown>>(children)) {
         return null
       }
+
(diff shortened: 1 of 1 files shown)


exKAZUu:
@segunadebayo Sorry, my PR contains type errors. I'm now fixing them. -> Done: #3563

PR #3589, opened 2025-08-15
feat: export utils

babakfp:
fixes https://github.com/chakra-ui/ark/discussions/3386

this exports all utils because I thought someone may find them useful.


segunadebayo:
@babakfp, what's the reasoning for this?

I'm weary of exposing utilities to reduce the exposed API surface.

babakfp:
`createSplitProps` is useful when creating custom components. Other exported utils may also be useful, so i exported them too.

cschroeter:
@babakfp
I’m open to exporting utilities like `createSplitProps`. In my opinion, the risk of exporting every utility and the additional maintenance and carefulness that would require from our side is simply not worth it.

That said, if we do this, we should also properly document every exported utility. If you’d still like to continue, please create a PR for each utility you want to expose, including documentation.

Alternatively, it might also be an option to just copy-paste the function you need, as they haven’t really changed in ages.

PR #3618, opened 2025-09-13
wrap ResizeObserver callback with requestAnimationFrame

IvanKalinin:
Since I've implemented a Frame feature in one of my projects, I get a lot of `Error ResizeObserver loop completed with undelivered notifications.` errors in BugSnag on the pages where I use the frame.
It is extremely hard to replicate, and I was not able to do that on my end. But the context points to the Frame component.
So, being honest, this is a blind attempt to resolve the issue, and according to my research, it appears to be a good practice to wrap `ResizeObserver` callback with `requestAnimationFrame` to prevent resize observer loops, which should not hurt anyway.


segunadebayo:
Thanks for this fix. It's a good practice in general to do so.

PR #3647, opened 2025-09-29
chore(svelte): improve portal

Hugos68:
This PR improves the Portal component in Svelte by reducing complexity and simplifying the $effect used.


Hugos68:
I ran `bun format` in the root and it seems that the entire repository wasn't properly formatted?

segunadebayo:
Thanks for attempting to improve Ark UI.

We used to have the Portal this simple in the past, but it caused some issues with SvelteKit. I prefer we keep it as is for now.

PR #3677, opened 2025-10-25
feat: image cropper

nelsonlaidev:


packages/solid/src/components/image-cropper/examples/basic.tsx
```diff
@@ -0,0 +1,26 @@
+import { For } from 'solid-js'
+import { ImageCropper } from '../'
```
segunadebayo: Examples should import from `@ark-ui/<framework>/<component>`

packages/solid/src/components/image-cropper/examples/basic.tsx
```diff
@@ -0,0 +1,26 @@
...
+    'top-left',
+    'top',
+    'top-right',
+    'right',
+    'bottom-right',
+    'bottom',
+    'bottom-left',
+    'left',
+  ] as const
+
```
segunadebayo: In the next zag version, we export the `imageCropper.handles` to replace this array

segunadebayo:
I've added a new grid part to the image cropper in the latest zag. Can you bump the versions and check?
https://github.com/chakra-ui/zag/blob/main/examples/next-ts/pages/image-cropper.tsx#L97

Update: I've bumped the zag version locally

nelsonlaidev:
@segunadebayo Thanks for refactoring the image cropper and adding the new grid part. The update is blocked by https://github.com/chakra-ui/zag/pull/2819 currently.

I suggest adding a script to check for any missing types in the machines. Based on my observations, all types should be re-exported except for `<machine>Schema`. But some machines exported `<machine>Schema as Schema`.

Should I go ahead to add it in the zag codebase?

nelsonlaidev:
Like:

CleanShot 2025-11-03 at 17 48 56@2x

segunadebayo:
@nelsonlaidev, Yes, please add that script; it'll be very useful.

PR #3743, opened 2025-12-29
feat(locale): add support to switch dir

AmeerDlshad:


AmeerDlshad:
Can we please merge this

segunadebayo:
After looking at how direction is handled in i18n, we’ve decided to **close this for now**.

Direction is treated as part of locale, not as a separate switch:

- In popular libraries like **next-intl** and the native **Intl** API derive direction from the locale (e.g., `Intl.Locale(locale).getTextInfo().direction`).
- In that model, `dir` is a consequence of the current locale, not an independent setting.

Our current behavior—`dir` derived from `locale` via `isRTL(locale)`—matches that: we don’t want to add a way to override `dir` at the provider level, since it would encourage decoupling direction from locale, which goes against how most intl libraries and the platform model it.

The main use case for a separate `dir` override would be “preview RTL with any locale” for QA/design. We’re not supporting that as a first-class feature in the locale provider right now, so we’d rather keep the API aligned with “dir comes from locale” only.

We’ll close this for now, but we’re happy to revisit if we later decide to support RTL-preview as an explicit

AmeerDlshad:
Ok then how am i supposed to make it so my app still shows English numbers while it's an RTL language like Arabic?

AmeerDlshad:
Or at least expose ```LocaleContextProvider``` to us from
```
var [LocaleContextProvider, useLocaleContext] = createContext({...})
```
so we can edit it without changing how ```LocaleProvider``` works.

PR #3811, opened 2026-03-10
fix(pin-input) PinInput crashes when the same value is entered in the same input

Hwacc:
This PR aims to fix issue [#3810 ](https://github.com/chakra-ui/zag/issues/3020)

Submitted `Vue` and `Solid` versions


segunadebayo:
This isn't the right fix. Will look into it

PR #3962, opened 2026-08-06
fix(react): fix compose refs

### LINKED ISSUE #3961: `composeRefs` doesn't reset all refs when composed with a React 19 cleanup-returning ref
### Description

`composeRefs` merges multiple refs (callback refs, object refs, and React 19 "cleanup ref" callbacks) into a single ref callback.

The composed callback only returns a dispose/cleanup function if **at least one** of the underlying refs returns a cleanup function (React 19 style). When it does, React switches to calling that single returned cleanup on unmount/detach — it no longer re-invokes the composed callback with `null` as it would for legacy refs.

The bug: the returned cleanup function only invoked the cleanups collected from refs that actually returned one. It never called plain callback refs with `null`, nor reset object refs' `.current` to `null`. So mixing a React 19 cleanup-ref with plain callback refs or object refs left those other refs stuck pointing at the detached/unmounted node.

See steps to reproduce

### Steps to Reproduce

```tsx
const refWithCleanup = (node) => {
  // ...setup
  return () => {
    // cleanup
  }
}
const plainCallbackRef = (node) => {
  // no return value
}
const objectRef = useRef(null)

const composed = useComposedRefs(refWithCleanup, plainCallbackRef, objectRef)

// mount, then unmount/detach `composed`
```

#### Expected result

`plainCallbackRef` called with `null`, `objectRef.current` === null

#### Actual result

neither happens, only `refWithCleanup`'s returned cleanup runs

### Ark UI Version

5.38.0

### Framework

- [x] React
- [ ] Solid
- [ ] Vue
- [ ] Svelte

### Browser

_No response_

### Additional Information

#3962

abarani:
Closes #3961 

## 📝 Description

`composeRefs` doesn't reset all refs when composed with a React 19 cleanup-returning ref.

Track whether any ref actually supplied a cleanup function. If none did, return undefined from the composed callback (preserving legacy behavior/back-compat with pre-React 19, where React re-invokes the same callback with null). If at least one ref did supply a cleanup, return a single dispose function that resets all refs: invoking real cleanups, calling plain callback refs with null, and nulling out object refs.

## ⛳️ Current behavior (updates)

Plain callback refs and object refs are not cleaned up, only `refWithCleanup`'s returned cleanup runs.

## 🚀 New behavior

Plain callback refs and object refs are not cleanup up correctly.

### DIFF
--- a/packages/react/src/utils/compose-refs.test.tsx
+++ b/packages/react/src/utils/compose-refs.test.tsx
@@ -15,15 +15,37 @@ describe('Util: composeRefs', () => {
     expect(objectRef.current).toBe(node)
   })
 
-  it('should run callback ref cleanups', () => {
+  it('should clean up mixed refs: cleanup callback, plain callback, and object ref', () => {
     const node = document.createElement('div')
     const cleanup = vi.fn()
-    const callbackRef = vi.fn(() => cleanup)
+    const refWithCleanup = vi.fn(() => cleanup)
+    const refWithoutCleanup = vi.fn()
+    const objectRef = { current: null as HTMLDivElement | null }
 
-    const dispose = composeRefs(callbackRef)(node) as VoidFunction | undefined
+    const dispose = composeRefs(refWithCleanup, refWithoutCleanup, objectRef)(node) as VoidFunction | undefined
+    expect(dispose).toBeTypeOf('function')
+
     dispose?.()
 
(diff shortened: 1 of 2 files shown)


segunadebayo:
Fixed directly on main in 1af4b0ac5, closing in favor of that fix.

PR #3987, opened 2026-08-20
feat: expose DOM attribute utilities

isBatak:
## Summary

- expose `ariaAttr` and `dataAttr` from each framework package root through Ark-owned `utils/attr.ts` modules
- use the local utility modules in existing field and fieldset implementations
- add patch changesets for React, Solid, Svelte, and Vue


Adebesin-Cell (approved):
Thank you! ❤️

PR #3994, opened 2026-08-31
fix(tour): keep the backdrop hidden after the tour closes

moczolaszlo:
## Problem

`Tour.Backdrop` applies `hidden` *after* merging the presence props, so it discards the `hidden` that presence itself provides:

```tsx
const mergedProps = mergeProps(tour.getBackdropProps(), presence.getPresenceProps(), props)
…
<ark.div {...mergedProps} ref={composedRefs} hidden={!tour.step?.backdrop} />
```

With the default render strategy (`unmountOnExit: false`) the node stays mounted after the tour closes, and presence sets `hidden: true` on it. That value is dropped, and `!tour.step?.backdrop` is `false` at that point, because the machine keeps `step` after a dismiss — `DISMISS` no longer runs `clearStep` as of `@zag-js/tour@1.43.3`. So the backdrop stays on screen for good, while the content disappears (the machine sets `hidden` on that one itself).

Reproducible on the docs site with any tour that has a backdrop: start it, close it, and the dimming remains. Measured on the closed backdrop: `hidden="false"`, `data-state="closed"`, `data-type="tooltip"` (the retained step), `display: block`.

Up to `@ark-ui/react@5.37.2` the same line was harmless: closing the tour cleared `step`, so `!tour.step?.backdrop` happened to be `true` and hid the node.

## Fix

Keep both reasons to hide, rather than only the step's:

```tsx
hidden={mergedProps.hidden || !tour.step?.backdrop}
```

Reordering the merge so presence wins outright would break the other direction: for a step with `backdrop: false`, presence's `hidden: false` would take over while the tour is open, and the backdrop would appear where it shouldn't.

Solid has the same shape and gets the same change. Vue is unaffected: it renders through `v-if` and binds the merged props, so presence's `hidden` isn't overridden.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/tour-backdrop-presence-hidden.md
+++ b/.changeset/tour-backdrop-presence-hidden.md
@@ -1,0 +1,6 @@
+---
+'@ark-ui/react': patch
+'@ark-ui/solid': patch
+---
+
+Fix issue where the tour backdrop stayed visible after the tour was closed
--- a/.changeset/tour-backdrop-presence-hidden.md
+++ b/.changeset/tour-backdrop-presence-hidden.md
@@ -1,6 +1,7 @@
 ---
 '@ark-ui/react': patch
 '@ark-ui/solid': patch
+'@ark-ui/svelte': patch
 ---
 
 Fix issue where the tour backdrop stayed visible after the tour was closed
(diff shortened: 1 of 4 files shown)


moczolaszlo:
Yes, the PR description is generated by Claude, I hope it's not a problem.

PR #4020, opened 2026-09-07
fix(vue): stop boolean part props defaulting to false

Adebesin-Cell:
## 📝 Description

Vue casts an absent boolean prop to `false` rather than leaving it `undefined`. zag therefore never saw these props as unset, and its own defaults never applied. Root components already guard against this with `BooleanDefaults`; the parts did not.

This turned up while root-causing the Vue `color-picker` test failure, which has been red long enough to get written off as a cross-framework parity issue. It wasn't — it's this.

## ⛳️ Current behavior

`Accordion.Root disabled` does not reach its items:

```
vue    disabled-attr=false
react  disabled-attr=true
```

The item sends `disabled: false`, and zag resolves `props.disabled ?? prop("disabled")`, so an explicit `false` beats the root's value. React sends `undefined` and inherits correctly.

`ColorPicker.Swatch` renders hex rather than the picker's format:

```
react  style="--color: hsla(300, 100%, 50%, 1); background: hsla(300, 100%, 50%, 1)"
vue    style="--color: #FF00FF; background: #FF00FF"
```

zag does `swatchState.value.toString(respectAlpha ? "css" : "hex")` with `const { respectAlpha = true } = props`. Vue delivers `respectAlpha: false`, so the `= true` default never fires and it picks `"hex"`.

`NavigationMenu.Link` does not close the menu on click, for the same reason with `closeOnClick`.

## 🚀 New behavior

16 parts now declare their boolean props through `withDefaults(..., { key: undefined })`, matching what 48 root components already do:

```ts
const props = withDefaults(defineProps<AccordionItemProps>(), {
  disabled: undefined,
} satisfies BooleanDefaults<ItemProps>)
```

The guard is checked against the **zag** props type, not Ark's — `BooleanDefaults<AccordionItemProps>` pulls in every `aria-*` boolean from `HTMLAttributes` and demands defaults for all of them.

I audited all 18 parts with optional booleans rather than only the three that visibly break. The other 13 pass `disabled` / `invalid` / `readOnly` where zag treats `false` and `undefined` alike today, so nothing changes for them now — but they are one upstream default away from the same bug.

**The Vue suite is fully green for the first time in a while: 40 files, 0 failures.** The `color-picker` "should render with default value" test passes without being touched.

## 🧩 Frameworks

- [ ] React
- [ ] Solid
- [ ] Svelte
- [x] Vue
- [ ] Not framework-specific

React, Solid and Svelte pass props through without boolean casting, so none of them have this.

## 💣 Is this a breaking change (Yes/No):

No. It restores the documented defaults.

## ✅ Checklist

- [x] Added a changeset for user-facing changes
- [x] Added or updated tests
- [ ] Added an example for any new prop or part
- [ ] Updated the docs

`accordion/tests/boolean-defaults.test.ts` pins the inheritance case. It fails on `v6` without this change and passes with it.

## 📝 Additional information

**Two parts are deliberately left out.** `date-picker-input` and `listbox-input` make `tsc` fail outright:

```
error TS2590: Expression produces a union type that is too complex to represent.
```

`withDefaults` is what triggers it, not the `satisfies` guard — dropping the guard doesn't help, and neither does spreading props at the call site. Clean `v6` has zero TS2590, so this is introduced by the fix, not pre-existing. `fixOnBlur` on `date-picker-input` defaults to `true` upstream and so is genuinely affected; `autoHighlight` on `listbox-input` has no zag default and is latent. Both want a different approach and are worth their own issue.

**Three parts keep `withDefaults` without the `satisfies` guard** — `combobox-item`, plus the two above had it removed before being reverted. `ItemProps["item"]` is `any`-typed, so `boolean extends any` holds and `BooleanDefaults` demands a default for `item` as well.

<details><summary><h3>Greptile Summary</h3></summary>

The PR restores Vue’s intended omitted-boolean semantics for non-root component parts so Zag defaults and inherited state can apply.
- Adds undefined defaults for optional boolean props across 16 Vue part components.
- Adds an Accordion regression test covering inherited disabled state.
- Documents the user-facing behavior change in a patch changeset.
</details>

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| packages/vue/src/components/accordion/accordion-item.vue | Adds an undefined default for the item’s optional disabled prop, allowing omitted values to inherit root state. |
| packages/vue/src/components/accordion/tests/boolean-defaults.test.ts | Adds regression coverage confirming that an item with no local disabled value inherits the root’s disabled state. |
| packages/vue/src/components/color-picker/color-picker-swatch.vue | Preserves undefined for an omitted respectAlpha prop so the upstream default controls color formatting. |
| packages/vue/src/components/navigation-menu/navigation-menu-link.vue | Preserves undefined for omitted closeOnClick and current props before forwarding them to Navigation Menu. |
| packages/vue/src/components/combobox/combobox-item.vue | Adds an undefined default for the optional persistFocus prop while retaining the component’s existing forwarding behavior. |

</details>

<sub>Reviews (2): Last reviewed commit: [82b0971](https://github.com/chakra-ui/ark/commit/82b097161860cb91f57d9bc26789db52bf36ea1c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61304091)</sub>


packages/vue/src/components/accordion/tests/boolean-defaults.test.ts
segunadebayo, line 1: Please don't use h for tests. Always use a Vue file for everything..testing and examples

Adebesin-Cell: Done — swapped to a `.vue` component in 82b0971.

Writing it as a Vue file turned out to matter for more than style. Vue only boolean-casts a prop that is genuinely absent from the vnode props, so `:disabled="undefined"` and omitting `disabled` behave differently:

```vue
<Accordion.Item :disabled="item.disabled" />   <!-- key present, stays undefined, inherits -->
<Accordion.Item value="react" />               <!-- key absent, cast to false, does not inherit -->
```

That is why `accordion.test.vue` never caught this — it always binds `:disabled`, so its "should disable all items when disabled is true" test passes even with the bug. The new `boolean-defaults.test.vue` omits the prop, which is what consumer code normally does.

Verified it reproduces: reverting just `accordion-item.vue` makes it fail with `expect(element).toBeDisabled()`, and it passes with the fix. Vue suite is 41 files, 0 failures.

PR #4031, opened 2026-09-09
feat(solid): forward part state into the render function

Adebesin-Cell:
## 📝 Description

Two fixes to Solid's `render`, which shipped in v6 without either.

**State.** `render` got the props to spread but no state, so its second argument was always the frozen empty object. 137 parts now forward the state their machine exposes.

**Merging.** `render` handed over a plain props object, so a caller adding its own handler replaced the part's rather than composing with it — a trigger given an `onClick` stopped toggling, silently. React avoids this through its element form, where the factory merges. Solid JSX evaluates eagerly and has no element to clone, so `asChild`'s callback was the only merge path and it is on its way out.

## ⛳️ Current behavior

```tsx
// state is EmptyState — state.open does not exist
<Collapsible.Trigger render={(props, state) => <button {...props}>{state.open ? 'Open' : 'Closed'}</button>} />

// the trigger stops toggling, with no warning
<Collapsible.Trigger render={(props) => <button {...props} onClick={mine} />} />
```

## 🚀 New behavior

`props` is now the same function `asChild` takes: call it to get the part's props, passing your own to merge them in. `state` is an accessor, since Solid never re-executes a component body and a by-value state would freeze at its initial value.

```tsx
<Collapsible.Trigger render={(props, state) => <button {...props()}>{state().open ? 'Open' : 'Closed'}</button>} />

<Collapsible.Trigger render={(props) => <button {...props({ class: 'mine', onClick: mine })} />} />
```

Props belonging to your component rather than the element stay outside the call: `{...props()} item={virtualItem()}`.

## 🧩 Frameworks

- [ ] React
- [x] Solid
- [ ] Svelte
- [ ] Vue

## 💣 Is this a breaking change (Yes/No):

Yes, twice — both within v6, so no released code breaks.

- `render`'s second argument was `EmptyState`, so nothing could read it.
- `{...props}` becomes `{...props()}`, and `render={MyButton}` is gone: a component reference would now receive the function rather than props. It had no callers outside its own test.

## ✅ Checklist

- [x] Added a changeset for user-facing changes
- [x] Added or updated tests
- [x] Added an example for any new prop or part
- [x] Updated the docs

## 📝 Additional information

**Not done here.** The per-part `State` types are declared but not exported — `scripts/exports-sync` copies React's barrels over Solid's and `check:exports` diffs react/solid/vue, so exporting from Solid alone fails CI. It needs React #4023 and Svelte #4030 first. The regenerated type tables name those types in the meantime.

**Also in here.** The type tables never documented `render` at all; regenerating them also picks up drift this branch did not cause (`onEnterComplete`, `popupType`), isolated in its own commit. `TreeViewNodeState` and `ScrollAreaScrollbarState` were dropped as duplicates of names the barrels already export. `Collapsible.Root` is generic over the state it forwards rather than `any`. `TreeView.NodeGroup` forwards `getNodeState`, matching `Accordion.Item`.

Typecheck and `check:exports` pass; Solid is 355 passed / 2 failed, the failures being the known `select` `translations` crash from zag next.2, unchanged from `v6`.


segunadebayo:
Don't merge this one. I'm reworking it to match the design intent

PR #4044, opened 2026-09-15
fix(svelte): keep the select value text's placeholder off the DOM

aaron-easygo:
## Description

`Select.ValueText` on the Svelte adapter spreads every prop onto the underlying `span`, including `placeholder` and `children`, which it also renders itself.

## Current behavior

```svelte
<Select.ValueText placeholder="Select" />
```

renders `<span ... placeholder="Select">Select</span>` — `placeholder` is not a valid attribute on a `span`, and the React adapter destructures it out, so the two stacks emit different markup for the same call.

## New behavior

`placeholder` and `children` are destructured before the merge, matching `Listbox.ValueText` and `DatePicker.ValueText`, which already do this.

Covered by a new case in `select.test.ts`; it fails on `main`.

## Is this a breaking change

No.

<h2><a href="https://app.greptile.com/api/retrigger?id=64291060"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge with no actionable correctness, security, or repository-rule issues identified.

<details><summary><h3>Summary</h3></summary>

This PR aligns Svelte’s `Select.ValueText` with equivalent value-text components by keeping `placeholder` and `children` out of the underlying span’s attributes.
- Destructures component-only props before merging DOM props.
- Continues rendering custom children or placeholder fallback text.
- Adds regression coverage and a patch changeset.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(svelte): keep the select value text&#39;..."](https://github.com/chakra-ui/ark/commit/44fa9198928c266d6a9cbb88b8b59184270e679e)</sub>


Adebesin-Cell (approved):
Cool!

PR #4090, opened 2026-09-19
fix(website): force-static docs pages so components prerender

Adebesin-Cell:
## 📝 Description

Fixes the docs site 500ing on component pages in production. `/docs/[...slug]` is opted into static generation so every page prerenders.

## ⛳️ Current behavior

The route has `generateStaticParams` returning all pages (confirmed: 54 components × 4 frameworks), but under the default `dynamic = 'auto'` Next prerendered only a couple of component pages (`fieldset`, `overview`) and served the rest **on demand** (`ƒ`). Those dynamically-rendered pages 500 in the Vercel runtime, so every component page (dialog, tabs, navigation-menu, …) returned a server error. It worked in `next dev` and even in a local `next start`, which is why it only showed up on the deploy.

## 🚀 New behavior

`export const dynamic = 'force-static'` on the docs route. All **216** component pages now prerender to static HTML (verified: `dialog.html`, `navigation-menu.html`, … exist; served with `x-nextjs-prerender: 1`, `x-nextjs-cache: HIT`, 200). No page needs per-request data — the content is fully static — so forcing static is safe and also makes the docs faster.

## 🧩 Frameworks

- [x] Not framework-specific

## ✅ Checklist

- [ ] Added a changeset for user-facing changes 
- [ ] Added or updated tests
- [ ] Added an example for any new prop or part
- [x] Updated the docs

## 📝 Additional information

Diagnosed by instrumenting `generateStaticParams` (it returns all 54 components) and rebuilding with `dynamic = 'error'`, which built cleanly and prerendered all 216 pages — proving the pages are static-capable and `auto` was needlessly demoting them. `force-static` is the safe production directive; `error` would also work and would additionally fail the build if a docs page ever became dynamic again.


Adebesin-Cell:
Closing in favor of #4091. `force-static` was a workaround

